PR #1, opened 2017-01-20
Update db schema to support 64 characters with device IDs.

novabyte:
In Unity `SystemInfo.deviceUniqueIdentifier` is not guaranteed to be a string with a fixed length. The length of the UDID depends on the platform which it's run on. Windows 10 appears to generate IDs which are 40 characters and exceed our input limit in the database.


mofirouz:
LGTM +1

PR #3, opened 2017-02-02
Add Debug build flag. 

mofirouz:
Changes in this PR:

- Add Debug build flag to enable debug logging and console output in development builds. This will disable processing of messages with `Zap.Level.Debug`. It will also enable mirroring log out to the console in development builds.

- Various query tweaks for Friends and Storage.

- Add additional log messages and enhancing a couple of existing ones.

- Fix issue where random handle generator wasn't seeded properly.

- Logout messages now close the connection as well and won't reply.

- Fix sending Close frame message in the Websocket in order to gracefully close connection.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,12 +44,13 @@ const (
 var (
 	version  string
 	commitID string
+	debugBuild bool = true
 )
 
 func main() {
 	semver := fmt.Sprintf("%s+%s", version, commitID)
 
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -44,12 +43,18 @@ const (
 var (
 	version  string
 	commitID string
+	debugBuild bool = true
```
novabyte: Let's convert this over to a `--verbose` flag in CLI arguments rather than as a build time constant.

novabyte:
@mofirouz Thanks. I'll merge it in. 👍

PR #4, opened 2017-02-06
Update Homebrew formula for stable builds.

novabyte:
This switches the brew formula over to use stable releases by default.

The `HEAD` builds do not work right now. I'll investigate further but it's related to the ephemeral `GOPATH` is setup by Homebrew for the build and the tools are not on the `PATH` when executed by the Makefile.


zyro:
@novabyte Ship it! 👍

PR #16, opened 2017-02-08
Change Friends to use Byte[] rather than String

mofirouz:
- Update TFriendAdd, TFriendRemove, TFriendBlock to accept UserID as bytes
- Tweak storage queries and fix verbose mode


novabyte:
@mofirouz Looks good 👍

PR #18, opened 2017-02-09
Let dbname be set in migrate and at server start.

novabyte:
- Various SQL schema improvements for wider compatibility.
- Let dbname be set in `nakama migrate` and `nakama` server start.

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,20 +20,22 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"syscall"
 	"time"
 
+	"nakama/cmd"
+	"nakama/pkg/ga"
+	"nakama/server"
+
 	"github.com/armon/go-metrics"
 	"github.com/go-yaml/yaml"
 	_ "github.com/lib/pq"
 	uuid "github.com/satori/go.uuid"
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -222,7 +223,17 @@ func configureLogger(clogger zap.Logger, config server.Config) (zap.Logger, zap.
...
 	// TODO config database pooling
-	db, err := sql.Open("postgres", "postgresql://"+dsns[0]+"/nakama?sslmode=disable")
+	rawurl := fmt.Sprintf("postgresql://%s?sslmode=disable", dsns[0])
+	url, err := url.Parse(rawurl)
+	if err != nil {
+		multiLogger.Fatal("Bad connection URL", zap.Error(err))
+	}
+
+	if len(url.Path) < 1 {
+		url.Path = "/nakama"
```
mofirouz: does this need to include the `/` - earlier on in the migration this wasn't needed?

mofirouz:
A small comment to double check whether a `/` prefix is needed or not. Otherwise LGTM 👍

main.go
novabyte: What line do you refer to earlier in the migration? Have a look at - https://github.com/heroiclabs/nakama/pull/18/files#diff-f1bf5486c350dca6f6f470fa77823555R136

PR #23, opened 2017-02-12
Run nakama (and cockroachdb) in Docker

mofirouz:
- Add `Dockerfile` to create a Nakama image in Docker.
- Add `docker-compose.yml` configuration to run Cockroach, Nakama Migration and Nakama as a whole setup. This enables a complete seamless developer experience. Developers can get started with Nakama using just one command:

     ```
     docker-compose up
     ```

- Both Docker and Cockroach are exposed to the host machine using their default ports.
- Nakama's Docker image can run on its own without Docker-Compose. This is useful if you'd need to run Nakama migration at some later point:

    ```
    docker run heroiclabs/nakama migrate up
    ```


novabyte:
@mofirouz Very nice. Docker FTW :godmode: 👍

Issue #22, opened 2017-02-11
Enhancements for checking database state when starting


mofirouz:
When starting a node we check if the database is migrated to the latest available schema. However, this is not fool-proof if the database is not fully available. 

https://github.com/heroiclabs/nakama/blob/master/cmd/migrate.go#L65

We need to return with an error and fail early here as it will cause problems down the line.

novabyte:
This was resolved in the [0.11.2 release](https://github.com/heroiclabs/nakama/releases/v0.11.2).

Issue #21, opened 2017-02-11
Print database DSNS when starting


mofirouz:
Currently we print a bunch of information on server start but we don't print the connection information to the database before attempting to connect. We should definitely do this both in main as well as when migrating.

mofirouz:
This is now done when we start nakama, but not yet done when we do migration. Commit 03b664d9112883a5d8fb69b86e0214feac3e2fc3

novabyte:
This was resolved in the [0.11.2](https://github.com/heroiclabs/nakama/releases/v0.11.2) release.

PR #30, opened 2017-02-17
Fail fast if cannot retrieve migration history from database

mofirouz:
- Fail fast if cannot retrieve migration history from database. #22 
- Print database DSNS when migrating. #21 
- Improve checking for existing database entity in the database.


novabyte:
@mofirouz Looks good 👍

PR #49, opened 2017-03-15
Addition of specific error messages to the Error payload

mofirouz:


novabyte:
@zyro Looks good 👍 /cc @mofirouz

PR #58, opened 2017-03-28
Rework embedded dashboard.

novabyte:
- Use [Mithril](https://mithril.js.org/) instead of [VueJS](https://vuejs.org/) as the embedded dashboard framework.
- Switch to [BlazeCSS](http://blazecss.com/) for the UI view.
- Remove source maps and other dev output bundled in releases.


novabyte:
I've not changed the UX at all in this work. There's only minor differences to the styles.

screen shot 2017-03-28 at 14 14 36

screen shot 2017-03-28 at 14 17 34

PR #63, opened 2017-04-22
Update health endpoint to be on the main server port. 

mofirouz:
- Update health endpoint to be on the main server port. 
- Streamline Docker releases


novabyte:
@mofirouz Looks good 👍

PR #68, opened 2017-05-15
Streamline commandline arguments and config file options

mofirouz:
- Command line options and configuration file have the same options.
- Move `verbose` and `logtostdout` to a new `Log` section in config. To set them, use:
```
--log.verbose = true
--log.stdout = true
```
- Imported two files from FlagOverride from https://github.com/uber-go/flagoverride. There are a few changes:
  - Fixed issues with any config slices throwing a NPE.
  - Add the ability to set a custom FlagSet. This allows one to add options to FlagSet that aren't available in the struct (in our case, just simply a `config` filepath option for obvious reasons). Also this allows us to name the flagset appropriately.
  - Add a new custom tag called `usage` so we can set usage information. In FO, the name was used as the usage string which is a little lame.

- Rename `ops` to `dashboard`


novabyte:
@mofirouz Looks good. 👍 

We'll need to remember to update documentation and I'll add some detailed notes on the way command flags have changed into our CHANGELOG.

PR #69, opened 2017-05-22
Matchmaking

zyro:
* New matchmaking feature.
* Optionally target match data to a subset of match participants.

### DIFF
--- a/server/api.proto
+++ b/server/api.proto
@@ -156,6 +156,11 @@ message Envelope {
     TLeaderboards leaderboards = 57;
     TLeaderboardRecord leaderboard_record = 58;
     TLeaderboardRecords leaderboard_records = 59;
+
+    TMatchmakingStart matchmaking_start = 60;
+    TMatchmakingCancel matchmaking_cancel = 61;
+    TMatchmakingTicket matchmaking_ticket = 62;
+    MatchmakingResult matchmaking_result = 63;
   }
 }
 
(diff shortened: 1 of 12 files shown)


server/api.proto
```diff
@@ -156,6 +156,11 @@ message Envelope {
     TLeaderboards leaderboards = 57;
     TLeaderboardRecord leaderboard_record = 58;
     TLeaderboardRecords leaderboard_records = 59;
+
+    TMatchmakingStart matchmaking_start = 60;
```
zyro: I prefer `TMatchmaking*` over `TMatchmake`, but I can get behind `Add` and `Remove` instead of `Start` and `Cancel` if you and @mofirouz feel it works better. `TMatchmakingMatched` (or `TMatchmakeMatched`) sounds a little awkward to me. @mofirouz wdyt?

mofirouz: I prefer `TMatchmake`

zyro: OK sounds like the vote says `TMatchmake*` so I'll rename everything in the server, then the client. @mofirouz what about the suffixes? `Start`/`Cancel`/`Ticket`/`Result` or `Add`/`Remove`/`Ticket`/`Matched` or some other combination?

mofirouz: I prefer Add/Remove.

mofirouz:
Looks good. +1

PR #73, opened 2017-06-01
Storage validation improvements

zyro:


server/core_storage.go
```diff
@@ -311,9 +310,15 @@ WHERE `
 		if len(key.UserId) != 0 {
 			if uid, err := uuid.FromBytes(key.UserId); err != nil {
 				return BAD_INPUT, errors.New("Invalid user ID")
+			} else if caller != uuid.Nil && caller != uid {
+				// If the caller is a client, only allow them to write their own data.
+				return BAD_INPUT, errors.New("Clients can only remove their own data")
```
novabyte: I think we should improve the error message to "A client can only remove user-owned records.".

server/core_storage.go
```diff
@@ -311,9 +310,15 @@ WHERE `
...
 			} else {
 				owner = uid.Bytes()
 			}
+		} else if caller != uuid.Nil {
+			// If the caller is a client, do not allow them to write global data.
+			return BAD_INPUT, errors.New("Clients cannot remove global data")
```
novabyte: Similar as above "A client cannot remove global records.".

PR #75, opened 2017-06-07
Runtime Base64 and Base16 conversion util functions.

mofirouz:


novabyte:
@mofirouz Looks good 👍 I'll merge it in

PR #77, opened 2017-06-14
Add batch api changes

mofirouz:


server/api.proto
```diff
@@ -261,18 +261,26 @@ message Friend {
 }
 
 message TFriendAdd {
+  message UserIds {
```
zyro: This means we can only add by IDs or by handles. How about using a message with a single `oneof` ID or handle, and repeating that? Does it make sense or would it be painful to handle server-side?

server/api.proto
```diff
@@ -401,32 +426,53 @@ message UserPresence {
...
+  }
+  message GroupId {
+    repeated bytes group_id = 1;
+  }
+
   oneof id {
-    bytes user_id = 1;
-    bytes room = 2;
-    bytes group_id = 3;
+    UserId user_id = 1;
```
zyro: Same as adding a friend, how about we have a message with a single `oneof` between user ID, room name, or group ID and repeating that instead. If would allow topic joins for any mix of DM, room, or group topics.

server/api.proto
```diff
@@ -480,33 +526,52 @@ message MatchmakeMatched {
...
+    repeated bytes match_id = 1;
+  }
+  message Token {
+    repeated bytes token = 2;
+  }
+
   oneof id {
-    bytes match_id = 1;
-    bytes token = 2;
+    MatchId match_id = 1;
```
zyro: Same as above.

server/api.proto
mofirouz: In other places, like `TMatchJoin` we used the approach in this PR. Changing it to the way you mentioned in the comment will make the API inconsistent unless we change those places as well. 

The server-side complexity is a little bit more, but not too hard.

I can be persuaded in either direction :)

server/core_user.go
```diff
@@ -126,3 +126,37 @@ func UsersFetchHandle(logger *zap.Logger, db *sql.DB, handles []string) ([]*User
...
+		statements = append(statements, statement)
+		params = append(params, handle)
+	}
+
+	query := ""
+	if len(userIds) > 0 {
+		query += " WHERE users.id IN (" + strings.Join(statements, ", ") + ") "
+	}
+	if len(handles) > 0 {
+		query += " WHERE users.handle IN (" + strings.Join(statements, ", ") + ") "
```
zyro: I believe this query won't work if you have both `userIds` and `handles`. It should be `WHERE ... OR ...`, and the handles should default to `WHERE` if there were no `userIds`.

server/pipeline_self.go
```diff
@@ -95,7 +95,7 @@ WHERE u.id = $1`,
 			LastOnlineAt: lastOnlineAt.Int64,
 		},
 		Email:        email.String,
-		DeviceId:     deviceIDs,
+		DeviceIds:     deviceIDs,
```
zyro: I think the spacing is off here.

PR #81, opened 2017-06-22
Storage list feature

zyro:
Query analysis results [here](https://gist.github.com/zyro/7026094f36851534ab8e2945601749d6).

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -35,13 +35,13 @@ type storageListCursor struct {
 	Read       int64
 }
 
-type StorageListOp struct {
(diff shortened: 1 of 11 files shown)


server/core_storage.go
```diff
@@ -22,10 +22,27 @@ import (
...
 
+type storageListCursor struct {
+	Bucket     string
+	Collection string
+	Record     string
+	UserID     []byte
+	Read       int64
+}
+
+//type StorageListOp struct {
```
novabyte: Do we need this code anymore?

PR #86, opened 2017-06-29
Lua proto message name lookup

mofirouz:


novabyte:
@mofirouz Looks good. Do we want to validate the message name in the Lua register functions? This way we can raise a WARN severity log message when a registration is made for a message type which doesn't exist in the server?

novabyte:
@mofirouz Nice 👍

PR #88, opened 2017-07-03
Ban users and create groups from within the Runtime environment.

mofirouz:
- Update Runtime modules to uses plural function names for batch operations. (`users_fetch_id` and `users_fetch_handle`.

- Ban users and create groups from within the Runtime environment.

- Also change `groupsCreate` message to run `TGroups` (rather than `TGroup`) since it's a batch operation. Therefore, removing `TGroup` message and reordering the proto messages once-again. **This requires a Client update.**


mofirouz:
@zyro Review plz :)

PR #89, opened 2017-07-09
Add In-App Purchase Validation

mofirouz:
This PR adds support for checking in-app purchase receipts against Apple and Google.

Few things to note:

- Apple and Google IAP remote validation is written as a library package. Therefore it's Nakama-agnostic for community's benefit.
- The actual Nakama ledger is therefore a separate process and is kept separated as it needs database schema support.

- The current design does not support multi in-app purchase receipts from Apple. It actually only processes the first item. This was a design decision.

- There are tests available for both the remote IAP validation library as well as tests that checks database interactions.


server/core_purchase.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		if err != sql.ErrNoRows {
+			r.Success = false
+			r.Message = errors.New("Failed to validate Apple purchase against ledger.")
+			p.logger.Error(r.Message.Error(), zap.Error(err))
+		}
+	}
+
+	// We've not seen this transaction
+	if purchaseUserID == nil || len(purchaseUserID) == 0 {
```
zyro: The length of a `nil` slice is 0, so this can be simplified to just `if len(purchaseUserID) == 0 {`.

server/core_purchase.go
```diff
@@ -0,0 +1,153 @@
...
+	if !r.Success {
+		return r
+	}
+
+	//TODO: Improvement - Process more than one in-app receipts
+	inAppReceipt := appleReceipt.InApp[0]
+	p.checkUser(userID, r, 1, inAppReceipt.TransactionID)
+
+	if r.Success && !r.SeenBefore {
+		err := p.savePurchase(userID.Bytes(), 1, inAppReceipt.ProductID, inAppReceipt.TransactionID, purchase.ReceiptData, r.Data)
```
zyro: If we couldn't save the purchase, should we fail? It could indicate general DB issues but also a race condition where multiple verifications for the same purchase are triggered concurrently and all of them return success even though only one can write to the ledger.

server/core_purchase.go
```diff
@@ -0,0 +1,153 @@
...
+func (p *PurchaseService) ValidateGooglePurchaseProduct(userID uuid.UUID, purchase *iap.GooglePurchase) *iap.PurchaseVerifyResponse {
+	r, _ := p.GoogleClient.VerifyProduct(purchase)
+	if !r.Success {
+		return r
+	}
+
+	p.checkUser(userID, r, 0, purchase.PurchaseToken)
+	if r.Success && !r.SeenBefore {
+		jsonPurchase, _ := json.Marshal(purchase)
+		err := p.savePurchase(userID.Bytes(), 0, purchase.ProductId, purchase.PurchaseToken, string(jsonPurchase), r.Data)
```
zyro: Same as above, should we fail?

server/core_purchase.go
```diff
@@ -0,0 +1,153 @@
...
+func (p *PurchaseService) ValidateGooglePurchaseSubscription(userID uuid.UUID, purchase *iap.GooglePurchase) *iap.PurchaseVerifyResponse {
+	r, _ := p.GoogleClient.VerifySubscription(purchase)
+	if !r.Success {
+		return r
+	}
+
+	p.checkUser(userID, r, 0, purchase.PurchaseToken)
+	if r.Success && !r.SeenBefore {
+		jsonPurchase, _ := json.Marshal(purchase)
+		err := p.savePurchase(userID.Bytes(), 0, purchase.ProductId, purchase.PurchaseToken, string(jsonPurchase), r.Data)
```
zyro: Same as above, should we fail?

server/core_purchase.go
mofirouz: Given that the `receipt_id` is part of the primary key, I don't think the race condition would occur here, correct?

The DB being problematic is still true, but IMO it shouldn't prevent the validation of the purchase from continuing.

@novabyte wdyt?

novabyte: @zyro I believe multiple writes over the same subset of the keyspace in cockroach will be coordinated via the same leaserholder (and therefore serialized as writes). I'll have to double check though.

@mofirouz Is there a way we could return an error which indicated the database server could not be reached with IAP validation and let the game client decide how to handle it?

zyro: `Success` has already been decided here, but we haven't updated the ledger yet. If the ledger update then fails because of a concurrent verification for the same purchase both will return success to the client, right?

PR #92, opened 2017-07-14
Update users from script runtime

zyro:


server/core_self.go
```diff
@@ -0,0 +1,147 @@
...
+	Fullname  string
+	Timezone  string
+	Location  string
+	Lang      string
+	Metadata  []byte
+	AvatarUrl string
+}
+
+func SelfUpdate(logger *zap.Logger, db *sql.DB, updates []*SelfUpdateOp) (Error_Code, error) {
+	if len(updates) == 0 {
```
mofirouz: This check should not be part of Core DB functions.

server/pipeline_self.go
```diff
@@ -109,77 +107,37 @@ WHERE u.id = $1`,
...
-			session.Send(ErrorMessage(envelope.CollationId, USER_HANDLE_INUSE, "Handle is in use"))
-		} else {
-			logger.Warn("Could not update user profile", zap.Error(err))
-			session.Send(ErrorMessageRuntimeException(envelope.CollationId, "Could not update user profile"))
-		}
-		return
-	} else if count, _ := res.RowsAffected(); count == 0 {
-		session.Send(ErrorMessageRuntimeException(envelope.CollationId, "Failed to update user profile"))
-		return
+		session.Send(ErrorMessage(envelope.CollationId, code, err.Error()))
```
mofirouz: We should `return` early here.

server/core_self.go
mofirouz: This func should assume that the data passed into it is already validated elsewhere.

server/runtime_nakama_module.go
```diff
@@ -295,6 +296,121 @@ func (n *NakamaModule) usersFetchHandle(l *lua.LState) int {
...
+
+	updates := make([]*SelfUpdateOp, 0)
+	for _, updateRaw := range updatesRaw {
+		if updateTable, ok := updateRaw.(map[string]interface{}); !ok {
+			l.ArgError(1, "expects a valid set of user updates")
+			return 0
+		} else {
+			update := &SelfUpdateOp{}
+			// Validate the user ID.
+			if u, ok := updateTable["UserId"]; ok {
```
mofirouz: While this is correct, I think there is a nicer, more concise way of writing this code using `Table.ForEach(func(k lua.LValue, v lua.LValue){}`

Also, I think using `lua.LVAsNumber` and `v.String()` is a nicer way than casting data.

mofirouz: Have a look at `nk.groupsCreate` for inspiration.

server/core_self.go
```diff
@@ -0,0 +1,143 @@
...
+	defer func() {
+		if err != nil {
+			if rollbackErr := tx.Rollback(); rollbackErr != nil { // don't override value of err
+				logger.Error("Could not update user profile, rollback error", zap.Error(rollbackErr))
+			}
+		} else {
+			if e := tx.Commit(); e != nil {
+				logger.Error("Could not update user profile, commit error", zap.Error(e))
+				code = RUNTIME_EXCEPTION
+				err = errors.New("Could not update user profile, commit error")
```
mofirouz: "Could not update user profile, commit error" -> "Could not update user profile"

server/runtime_nakama_module.go
```diff
@@ -298,6 +299,109 @@ func (n *NakamaModule) usersFetchHandle(l *lua.LState) int {
...
+				}
+				if uid, err := uuid.FromString(v.String()); err != nil {
+					conversionError = "expects valid user IDs in each update"
+					return
+				} else {
+					update.UserId = uid.Bytes()
+				}
+			case "Handle":
+				if v.Type() != lua.LTString {
+					conversionError = "expects valid handles in each update, when provided"
```
mofirouz: Let's remove the `, when provided` suffix. This error only occurs when they have provided the Handle, can't occur in any other condition.

PR #95, opened 2017-07-18
Update error codes for login/register/link scenarios

zyro:


CHANGELOG.md
```diff
@@ -16,11 +16,14 @@ The format is based on [keep a changelog](http://keepachangelog.com/) and this p
 - Streamline command line flags to be inline with the config file.
 - Restructure and stabilize API messages.
 - Update Runtime modules to use plural function names for batch operations. (`users_fetch_id` and `users_fetch_handle`)
+- Attempting to login with an ID that does not exist will return a new dedicated error code.
+- Attempting to register with an ID that already exists will return a new dedicated error code.
```
mofirouz: Could you add the changes from the previous PR to here as well plz?

server/session_auth.go
```diff
@@ -242,23 +242,23 @@ func (a *authenticationService) StartServer(logger *zap.Logger) {
 }
 
 func (a *authenticationService) handleAuth(w http.ResponseWriter, r *http.Request,
-	retrieveUserID func(authReq *AuthenticateRequest) ([]byte, string, string, int)) {
+	retrieveUserID func(authReq *AuthenticateRequest) ([]byte, string, string, Error_Code, int)) {
```
mofirouz: Do we need to return the error code number (400, 401, 500) etc given that we have an `Error_Code` that we can `switch` on to produce the correct HTTP error number?

PR #96, opened 2017-07-19
Updated config and notification

mofirouz:


server/runtime_nakama_module.go
```diff
@@ -943,20 +943,21 @@ func (n *NakamaModule) groupsCreate(l *lua.LState) int {
...
-					l.ArgError(1, "expects Metadata to be string")
+					l.ArgError(1, "expects Metadata to be a table")
 					return
 				}
 
-				j := []byte(v.String())
-				maybeJson := make(map[string]interface{})
-				if jsonErr := json.Unmarshal(j, &maybeJson); jsonErr != nil {
+				metadataMap := ConvertLuaTable(v.(*lua.LTable))
+				metadataBytes, err := json.Marshal(metadataMap)
```
zyro: Can we avoid this conversion somehow? Perhaps make the core functions take maps as input, and move the input validation and unmarshal step into the pipeline when coming from the client?

server/runtime_nakama_module.go
```diff
@@ -1046,18 +1047,21 @@ func (n *NakamaModule) notificationsSendId(l *lua.LState) int {
...
+					l.ArgError(1, "expects Content to be a table")
 					return
 				}
-				j := []byte(v.String())
-				maybeJson := make(map[string]interface{})
-				if jsonErr := json.Unmarshal(j, &maybeJson); jsonErr != nil {
-					l.ArgError(1, "expects Content to be a valid JSON")
+
+				contentMap := ConvertLuaTable(v.(*lua.LTable))
+				contentBytes, err := json.Marshal(contentMap)
```
zyro: Same as above.

cmd/admin.go
```diff
@@ -59,7 +59,7 @@ func createLeaderboard(args []string, logger *zap.Logger) {
 	var metadata string
 
 	flags := flag.NewFlagSet("admin", flag.ExitOnError)
-	flags.StringVar(&dsns, "db", "root@localhost:26257", "CockroachDB JDBC connection details.")
+	flags.StringVar(&dsns, "database.address", "root@localhost:26257", "Address of CockroachDB server (username:password@address:port/dbname)")
```
zyro: If we're renaming the input, let's also change the `dsns` variable name to stay consistent.

cmd/migrate.go
```diff
@@ -240,7 +240,7 @@ func (ms *migrationService) status() {
 
 func (ms *migrationService) parseSubcommand(args []string) {
 	flags := flag.NewFlagSet("migrate", flag.ExitOnError)
-	flags.StringVar(&ms.DSNS, "db", "root@localhost:26257", "CockroachDB JDBC connection details.")
+	flags.StringVar(&ms.DSNS, "database.address", "root@localhost:26257", "Address of CockroachDB server (username:password@address:port/dbname)")
```
zyro: If we're renaming the input, let's also change the `ms.DSNS` field name to stay consistent.

zyro:
Can we apply the same table conversion to `users_update`'s metadata input?

cmd/migrate.go
mofirouz: What do you mean by this?

server/runtime_nakama_module.go
mofirouz: I don't like this. For instance, in the case of GroupCreate, the proto already expects the `metadata` field to be Bytes. No point putting this extra conversion in the pipeline.

server/runtime_nakama_module.go
mofirouz: There is no conversion happening anywhere else, because sending Notifications can only happen via Lua.

cmd/admin.go
mofirouz: Where do we have `dsns` ?

zyro: `flags.StringVar(&dsns, ...`

server/runtime_nakama_module.go
zyro: I wanted to avoid going from Lua to Go to JSON to Go, where the last step is the JSON validation. This seems to be moot as the group create core code isn't shared between the script runtime and the pipeline yet.

server/runtime_nakama_module.go
zyro: OK ignore this.

cmd/migrate.go
zyro: Do we still want to use the "DSNS" naming?

mofirouz:
@zyro please re-review :)

PR #98, opened 2017-07-23
Let config argument to be passed in any order

mofirouz:
- Invalid config file, or invalid command line config option results in server from starting. (FATAL error).
- Config file can be passed in any order, no longer requiring the `--config` param to be the first item.


zyro:
@mofirouz Looks good. :+1:

PR #99, opened 2017-07-24
Storage partial update feature

zyro:


server/core_storage.go
```diff
@@ -488,6 +505,164 @@ DO UPDATE SET value = $6::BYTEA, version = $7, read = $8, write = $9, updated_at
...
+		// Query and decode the row.
+		var userID []byte
+		var bucket sql.NullString
+		var collection sql.NullString
+		var record sql.NullString
+		var value []byte
+		var version []byte
+		err = tx.QueryRow(query, update.key.Bucket, update.key.Collection, owner, update.key.Record).
+			Scan(&userID, &bucket, &collection, &record, &value, &version)
+		if err != nil && err != sql.ErrNoRows {
```
mofirouz: What happens if no rows were returned?

server/core_storage.go
```diff
@@ -488,6 +505,164 @@ DO UPDATE SET value = $6::BYTEA, version = $7, read = $8, write = $9, updated_at
...
+		}
+		newVersion := []byte(fmt.Sprintf("%x", sha256.Sum256(newValue)))
+
+		query = `
+INSERT INTO storage (id, user_id, bucket, collection, record, value, version, read, write, created_at, updated_at, deleted_at)
+SELECT $1, $2, $3, $4, $5, $6::BYTEA, $7, $8, $9, $10, $10, 0`
+		params := []interface{}{uuid.NewV4().Bytes(), owner, update.key.Bucket, update.key.Collection, update.key.Record, newValue, newVersion, update.permissionRead, update.permissionWrite, ts}
+		if len(version) == 0 {
+			// Treat this as an if-none-match.
+			query += " WHERE NOT EXISTS (SELECT record FROM storage WHERE user_id = $2 AND bucket = $3 AND collection = $4 AND record = $5 AND deleted_at = 0)"
```
mofirouz: Do we care for this? Version was already not set, meaning data was not found and it's a new record that we are storing, so no need for WHERE NOT EXISTS?

-- 

Also, for an actual "if-none-match" case, shall we not check if version == "*" ?

server/core_storage.go
zyro: Line 594 the record is initalized as a new value.

server/core_storage.go
zyro: I'll double-check the OCC cases and tighten them up.

PR #102, opened 2017-07-28
Group function updates

zyro:
* Add script runtime function to update groups.
* Add script runtime function to list groups a user is part of.
* Add script runtime function to list users belonging to a group.
* Group leave operations now return a specific error code when the last admin attempts to leave.
* Group self list operations now return the user's membership type to each group.
* Remove `admin` sub-command.


server/api.proto
```diff
@@ -630,6 +633,20 @@ message TGroups {
   bytes cursor = 2;
 }
 
+message TGroupsSelf {
```
mofirouz: Missing documentation

PR #103, opened 2017-07-31
Notification wiring

zyro:


server/core_friend.go
```diff
@@ -18,29 +18,73 @@ import (
...
+			}
+			var recipient []byte
+			var sender []byte
+			var subject string
+			var code int64
+			if isFriendAccept {
+				recipient = userID
+				sender = friendID
+				subject = fmt.Sprintf("%v accepted your friend request", handle)
+				code = 3
```
mofirouz: Can we use some sort of an Enum here to indicate the message code, so it's not a magic number?

server/pipeline_friend.go
```diff
@@ -100,7 +103,39 @@ func (p *pipeline) addFacebookFriends(logger *zap.Logger, userID []byte, accessT
 				if err != nil {
 					logger.Error("Could not commit transaction", zap.Error(err))
 				} else {
-					logger.Info("Imported friends from Facebook")
+					logger.Debug("Imported friends from Facebook")
+
+					// Send out notifications.
+					if len(friendUserIDs) != 0 {
+						content, err := json.Marshal(map[string]interface{}{"handle": handle, "facebook_id": fbid})
```
mofirouz: Should this also include the user_id as well? I know they can query the accounts by handle, I'm just playing devil's advocate here

zyro: The sender always contains the User ID anyway, so no need to add it to the content. The only things added to the content are data fields that don't already appear in the top-level notification fields or are embedded in the subject: "foo wants to be your friend" -> content is `{"handle": "foo"}`. Enough information to display a custom/localised notification as needed.

server/core_friend.go
zyro: We might be able to use one internally but the benefit will be minimal as we still need to allow arbitrary custom codes from developers.

mofirouz: I think this is importantly for maintainability. Let's add this Enum in.

server/pipeline_friend.go
mofirouz: Fair enough

PR #107, opened 2017-08-04
Improve runtime before hook payload formats

zyro:


server/runtime.go
```diff
@@ -222,7 +225,35 @@ func (r *Runtime) InvokeFunctionRPC(fn *lua.LFunction, uid uuid.UUID, handle str
...
+		}
+	}
+
+	retValue, err := r.invokeFunction(l, fn, ctx, lv)
+	if err != nil {
+		return nil, err
+	}
+
+	if retValue == nil || retValue == lua.LNil {
+		return nil, nil
```
mofirouz: I don't think this is/should be valid return for a Before hook. They must return the envelop or raise an error if they want to stop pipeline processing.

server/runtime.go
```diff
@@ -312,3 +343,285 @@ func (r *Runtime) invokeFunction(l *lua.LState, fn *lua.LFunction, ctx *lua.LTab
...
+			return nil, errors.New("Invalid payload in GroupsLeave conversion from script runtime")
+		}
+		groupIds := groupsLeave.(*lua.LTable).RawGetString("GroupIds")
+		if groupIds.Type() != lua.LTTable {
+			return nil, errors.New("Invalid payload in GroupsLeave conversion from script runtime")
+		}
+		ids := make([][]byte, 0)
+		var err error
+		groupIds.(*lua.LTable).ForEach(func(k lua.LValue, v lua.LValue) {
+			if v.Type() != lua.LTString {
```
mofirouz: I think this check can be removed since `uuid.FromString` will produce an error if `v` is not a valid UUID string.

zyro: You're right, but this is consistent with our existing validation around ID fields. I'd rather keep them all consistent for now and drop the all the additional checks at once later.

PR #109, opened 2017-08-24
Sa core user.go update

sakiraykurt:
Our goal is to get social network ids in server side and to use them with third party services.
It is almost like client self fetch but no device ids only


mofirouz:
Thanks for this. This is fixed in the next release of the server. Closing this in favour of that.

PR #110, opened 2017-09-03
Replace Glide with Dep. Update to Zap 1.6.0

mofirouz:


zyro:
The readme still lists `glide` as a required tool, let's switch that to `dep`. Otherwise looks good 👍

PR #113, opened 2017-09-22
Clean up how runtime HTTP request headers are handled.

JoshuaStrunk:
This hopes to bring in line runtime HTTP request handling with auth HTTP request handling by running through media type headers through mime.ParseMediaType before evaluating them. It also works to improve communication of errors to developers by updating several error responses with both received and expected values.


zyro:
Thanks! I'll add some extra error handling on top of this now.

Issue #87, opened 2017-07-01
Matchmaker with metadata


ArisTsevrenis:
Be able to enter matchmaking with data other user count. 

Having a MatchOpCode would be benefitial for example and a potentially easy solution but generally having "rules" to join would be good.

nirolfa:
yes, custom matchmaking is needed -> shoud add some filter feature
(ex: match with only players between level 5 & 10 )

AlimJaffer:
Hi @nirolfa , can you give me some further details on what you’re looking for in the matchmaker?  What is your specific use-case?

prodigga:
Dont forget about https://github.com/heroiclabs/nakama/issues/11

The matchmaker really needs some love! :)

novabyte:
@prodigga It's not been forgotten just keep an eye out on the issue tracker. 😉

mofirouz:
This is now available on commit https://github.com/heroiclabs/nakama/commit/dc0a5964d8a1fe3dde79b1c6d1d8d0a59b71e216 and will be available in Nakama 1.1.0+

Issue #11, opened 2017-02-07
Enhance matchmaker sub-system


novabyte:
Support more complex fuzzy matches and multi-index search queries on user accounts and storage objects to generate opponent lists for multiplayer. This system will combine the in-memory presence system more directly with the persisted state to create more powerful query/filter composition. This depends on GIN indexes and full-text support on cockroachdb.

prodigga:
Maybe the API could be something like this:

```
var matchFindRequest = new NMatchFindMessage.Builder().
SearchRooms(string[]{"1v1","CaptureTheFlag","SearchAndDestroy"}).
AddSearchParam("MMR",50, 25).
SetTimeout(20).
Build();

client.Send(matchFindRequest , (INMatchMakingResult result) =>
{
  //found match
}, (INError error) => {
  //error
});
```

**_SearchRooms_** - Only match make with people in the same 'room'. This is just a way for us developers to filter out which kind of games the player is going to play in. Users do not 'create' rooms. Maybe developers create 'rooms' via command line on the server, similar to leaderborads? The developer can specify how many users per match in the room, etc. 

In the game, player presses: _Start -> Multiplayer -> Free For All -> Find Match_ so we send an _NMatchFindMessage_ with `SearchRooms(string[]{"FFA"})`. If user selects _Start -> Multiplayer -> All-> Find Match_, we send an _NMatchFindMessage_ with `SearchRooms(string[]{"FFA","DM","CTF"})`. 

_**AddSearchParam**_ - 'fuzzy' match making values. IE Pass in an "MMR" of "50"and a max search range of "25" (ie match against others with an MMR of 25-75. I am not sure if this is the 'best' way to do it, but just an idea.

_**SetTimeout**_ Time out for the search operation. Return an error if match not found after this time?

_**INMatchMakingResult **_ Contains a reference to the INMatch that was found, and also 

```
var cancelMatchFind = new NCancelMatchFindMessage.Default();
client.Send(cancelMatchFind, //etc....
```

To cancel current match find operation?

Just suggestions

prodigga:
Matchmaking should support group play. 

IE I want to play a match together with my friend. I ask nakama to match make for me and my friend (maybe I provide both our UserIDs as well as our "MMR"). The server puts us together in to a match.

I guess it would also be important that the players be out into the same team, so the server not only has to find a game with 2 free spots, but one with 2 free spots on the same team.

zyro:
[...]

PR #118, opened 2017-10-05
Add advanced matchmaker.

mofirouz:
- Update docker-compose to Cockroach 1.0.6

Using advanced matchmaker, you can:
- Setup single/multi-term filters, and setup how those filters should be matched (any of the terms, or all the terms must match)
- Matchmake based on a range
- Matchmake based on a boolean condition.
- You can setup as many filter as you like on a matchmaking request.

Alongside filters, you can supply user properties. Filters are matched against user properties, so the same filter-name and property-name must match before the filter is evaluated against that user.

You can supply extra user properties for metadata information. All user properties and all filters are redistributed back to each client when matches are found.

This answers #11 and #87.

### DIFF
--- a/server/api.proto
+++ b/server/api.proto
@@ -888,17 +888,17 @@ message TopicPresence {
 }
 
 /**
- * PropertyMap is a core domain type respresenting a single user property
+ * PropertyPair is a core domain type respresenting a single user property
  */
 message PropertyPair {
-  /// List of string user property
-  message StringList {
+  /// Set of string user property
+  message StringSet {
     repeated string values = 1;
   }
 
   string key = 1;
   oneof value {
-    StringList stringList = 2;
+    StringSet stringSet = 2;
     bool boolValue = 3;
     int64 intValue = 4;
   }
--- a/server/api.proto
+++ b/server/api.proto
@@ -888,13 +888,58 @@ message TopicPresence {
 }
 
 /**
+ * PropertyMap is a core domain type respresenting a single user property
+ */
+message PropertyPair {
+  /// List of string user property
+  message StringList {
+    repeated string values = 1;
+  }
+
+  string key = 1;
+  oneof value {
+    StringList stringList = 2;
+    bool boolValue = 3;
+    int64 intValue = 4;
+  }
+}
+
+/**
+ * MatchmakeFilter is a core domain type respresenting a filter to use for matchmaking.
+ */
+message MatchmakeFilter {
+  /// String term filters
+  message TermFilter {
+    repeated string terms = 1;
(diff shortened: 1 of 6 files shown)


server/api.proto
```diff
@@ -888,13 +888,58 @@ message TopicPresence {
 }
 
 /**
+ * PropertyMap is a core domain type respresenting a single user property
```
zyro: The comment references `PropertyMap` but the message name is `PropertyPair`, I think there was a change and these are no longer in sync?

server/matchmaker.go
```diff
@@ -51,36 +98,159 @@ func NewMatchmakerService(name string) *MatchmakerService {
...
+		}
 	}
-	if int64(len(selected)) == requiredCount-1 {
-		for mk, _ := range selected {
-			delete(m.values, mk)
+	return nil
+}
+
+func (m *MatchmakerService) checkFilter(requestProfile, queuedProfile *MatchmakerProfile) bool {
+	if queuedProfile.RequiredCount != requestProfile.RequiredCount {
```
zyro: Is GitHub messing up the formatting here, or is indent only one space?

server/matchmaker.go
```diff
@@ -51,36 +98,159 @@ func NewMatchmakerService(name string) *MatchmakerService {
...
+
+	return props
+}
+
+func (m *MatchmakerService) intersection(a, b []string) []string {
+	o := make([]string, 0)
+	for i := range a {
+		for j := range b {
+			if a[i] == b[j] {
+				o = append(o, a[i])
```
zyro: This algo is fine for now, but we can at least break out of the inner loop once we've found our item.

server/pipeline_matchmake.go
```diff
@@ -55,10 +87,56 @@ func (p *pipeline) matchmakeAdd(logger *zap.Logger, session *session, envelope *
...
+			case RANGE:
+				f := userFilterValue.(*MatchmakerRangeFilter)
+				filter.Value = &MatchmakeFilter_Range{&MatchmakeFilter_RangeFilter{f.LowerBound, f.UpperBound}}
+			case BOOL:
+				f := userFilterValue.(*MatchmakerBoolFilter)
+				filter.Value = &MatchmakeFilter_Check{f.Value}
+			}
+		}
+
+		logger.Info("propProp", zap.Any("pp", protoProp))
```
zyro: If we log these at all they should be at most `Debug` level I think.

server/matchmaker.go
zyro: Looks like GitHub compressing the indenting, never mind.

PR #119, opened 2017-10-15
Use separate runime underlying states to improve concurrency

zyro:


server/runtime.go
```diff
@@ -105,16 +98,91 @@ func NewRuntime(logger *zap.Logger, multiLogger *zap.Logger, db *sql.DB, config
...
+					vm:     vm,
+					luaEnv: ConvertMap(vm, config.Environment),
+				}
+
+				if err = r.loadModules(modules); err != nil {
+					multiLogger.Fatal("Failed initializing runtime modules", zap.Error(err))
+				}
+				return r
+
+				// TODO find a way to run r.Stop() when the pool discards this runtime.
```
mofirouz: Is this still a valid TODO? Do we need to do something about this?

zyro: @mofirouz It is valid, but not critical because it only explicitly cleans up temp files if any. We don't expose any Lua stdlib functions that access temp files.

There's no way to hook into `sync.Pool`'s eviction process so the only way to fully solve this is to use a different pool implementation.

PR #145, opened 2018-02-03
Wire up runtime scripts

zyro:


server/runtime_nakama_module.go
```diff
@@ -0,0 +1,1108 @@
...
+		l.ArgError(3, conversionError)
+		return 0
+	}
+
+	meta := n.tracker.GetLocalBySessionIDStreamUserID(sessionID, stream, userID)
+	if meta == nil {
+		l.Push(lua.LNil)
+	} else {
+		metaTable := l.NewTable()
+		metaTable.RawSetString("Hidden", lua.LBool(meta.Hidden))
```
mofirouz: Would be good to have a discussion about the Lua naming scheme - I know that we are currently using CamelCase for variable names and snake_case for func names - maybe a good time to consolidate the naming scheme. 

This also applies to Context variable names.

server/runtime_nakama_module.go
```diff
@@ -0,0 +1,1108 @@
...
+			}
+			stream.Label = v.String()
+		}
+	})
+	if conversionError != "" {
+		l.ArgError(3, conversionError)
+		return 0
+	}
+
+	meta := n.tracker.GetLocalBySessionIDStreamUserID(sessionID, stream, userID)
```
mofirouz: Does this code need changing in NE version? Different tracker call?

mofirouz: If so, maybe worth having a func call that looks at both Local and cluster presences?

server/runtime_nakama_module.go
```diff
@@ -0,0 +1,1108 @@
...
+		return 0
+	}
+
+	// By default generate presence events.
+	hidden := l.OptBool(4, false)
+	// By default persistence is enabled, if the stream supports it.
+	persistence := l.OptBool(5, true)
+
+	// Look up the session.
+	session := n.registry.Get(sessionID)
```
mofirouz: This only looks up the session on the local node, not cluster wide. I think we should have an offline call to figure out what the best thing here is.

server/runtime_nakama_module.go
```diff
@@ -0,0 +1,1108 @@
...
+}
+
+type NakamaModule struct {
+	logger              *zap.Logger
+	db                  *sql.DB
+	config              Config
+	registry            *SessionRegistry
+	tracker             Tracker
+	router              MessageRouter
+	once                *sync.Once
```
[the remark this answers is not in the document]
zyro: Will make sure the `nk.run_once` function isn't loaded until it's discussed more.

server/runtime_nakama_module.go
```diff
@@ -0,0 +1,1108 @@
...
+		"json_decode":          n.jsonDecode,
+		"base64_encode":        n.base64Encode,
+		"base64_decode":        n.base64Decode,
+		"base16_encode":        n.base16Encode,
+		"base16_decode":        n.base16decode,
+		"aes128_encrypt":       n.aes128encrypt,
+		"aes128_decrypt":       n.aes128decrypt,
+		"bcrypt_hash":          n.bcryptHash,
+		"bcrypt_compare":       n.bcryptCompare,
+		"generate_auth_token":  n.generateAuthToken,
```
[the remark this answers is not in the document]
zyro: Will do.

server/runtime_nakama_module.go
zyro: Stream presences are checked/mutated on the node they belong to only, so this is correct behaviour until that changes.

server/runtime_nakama_module.go
zyro: As above, stream presences are checked/mutated on the node they belong to only, so this is correct behaviour until that changes.

server/runtime_nakama_module.go
zyro: To be determined when we wire up conversion of data, especially for before/after hooks.

PR #153, opened 2018-02-13
Get account and update account

mofirouz:
- Remove `utc_offset_ms`
- Change `lang` to `lang_tag` and remove the NOT NULL constraint.
- Change `fullname` to `display_name`
- Change `created_at`/`updated_at`/`verified_at`/`disabled_at` to `create_time`, `update_time`,`verify_time`, `disable_time` in the database schema to map to the proto definitions.

- Add get account.
- Add update account.
- Add batch get users by ID, username or facebook ID.
- Fix device authentication problem, where if the device ID was already in the system and 'create' was set to True, it would fail to create a token.

@zyro A todo item for us to tackle later is to wire up the Online presence in the GetAccount call.


novabyte:
@mofirouz Looks good 👍

PR #158, opened 2018-02-20
Use Packr instead of go-bindata

mofirouz:
- `packr -z` only compresses the files. Doesn't invoke the Go build toolchain.
- Due to limitations in how Packr works (file naming scheme + file location) I am forced to move some files around.
- Packr uses the filesystem in local development, so checking in the generated file is not entirely necessary. I'm happy for us to check-in or add it to .gitignore.


migrations/migrate.go
```diff
@@ -49,11 +48,15 @@ type migrationService struct {
 	db         *sql.DB
 }
 
-func MigrationStartupCheck(logger *zap.Logger, db *sql.DB) {
+func StartupCheck(logger *zap.Logger, db *sql.DB) {
 	migrate.SetTable(migrationTable)
+
+	migrationBox := packr.NewBox("./sql") // path must be string not a variable for packr to understand
```
zyro: If we depend on it in code, should we include `packr` as a vendored dependency?

PR #159, opened 2018-02-21
Refactor accounts and users. Add notification proto definition.

mofirouz:
- Split users and account code into api_ and core_.
- Add notification proto.
- Add notification schema. Remove deleted_at and expires_at.


api/api.proto
```diff
@@ -617,10 +639,38 @@ message LinkFacebookRequest {
...
-// A collection of zero or more friends of the user.
-message Friends {
-  // The Friend objects.
-  repeated Friend friends = 1;
+// Fetch a list of unexpired notifications.
+message ListNotificationsRequest {
+  // Max number of notifications to list. Between 1 and 100.
+  int64 limit = 1;
+  // Use this cursor to paginate notifications. Cache this to catch up to new notifications. The value of this comes from NotificationList.cachable_cursor.
+  string cachable_cursor = 2;
```
zyro: Let's use `cacheable` as the spelling, seems to be the more standard form used in W3C texts etc

api/api.proto
```diff
@@ -617,10 +639,38 @@ message LinkFacebookRequest {
...
+  // True if this notification was persisted to the database.
+  bool persistent = 7;
+}
+
+// A collection of zero or more notifications.
+message NotificationList {
+  // Collection of notifications.
+  repeated Notification notifications = 1;
+  // Use this cursor to paginate notifications. Cache this to catch up to new notifications.
+  string cachable_cursor = 2;
```
zyro: Same as above.

migrations/sql/20180103142001_initial_schema.sql
```diff
@@ -63,7 +63,22 @@ CREATE TABLE IF NOT EXISTS user_edge (
...
+    user_id         UUID            NOT NULL,
+    subject         VARCHAR(255)    NOT NULL,
+    content         BYTEA           DEFAULT '{}' CHECK (length(content) < 16000) NOT NULL,
+    code            SMALLINT        NOT NULL,      -- Negative values are System reserved.
+    sender_id       UUID,                          -- NULL for System messages
+    create_time     BIGINT          CHECK (create_time > 0) NOT NULL,
+);
+
+-- list notifications for a user that are not deleted or expired, starting from a given ID (cursor).
+CREATE INDEX IF NOT EXISTS notification_user_id_deleted_at_expires_at_id_idx ON notification (user_id, deleted_at ASC, expires_at ASC, id);
```
zyro: I think we don't have `expires_at` anymore, right?

migrations/sql/20180103142001_initial_schema.sql
```diff
@@ -63,7 +63,22 @@ CREATE TABLE IF NOT EXISTS user_edge (
...
+
 -- +migrate Down
 DROP TABLE IF EXISTS user_edge;
 DROP TABLE IF EXISTS user_device;
 DROP TABLE IF EXISTS users;
+DROP TABLE IF EXISTS notification;
```
zyro: Let's drop in reverse order to creation. (Move the notification table drop to the top in this case.)

PR #160, opened 2018-02-22
Add in-app notifications

mofirouz:


server/core_friend.go
```diff
@@ -103,37 +111,47 @@ func AddFriends(logger *zap.Logger, db *sql.DB, currentUser uuid.UUID, ids []str
 		return err
 	}
 
-	// TODO(mo, zyro): Use notificationToSend to send notification here.
+	notifications := make(map[uuid.UUID][]*api.Notification)
+	for id, isFriendAccept := range notificationToSend {
+		uid := uuid.FromStringOrNil(id)
+		content, _ := json.Marshal(map[string]interface{}{"username": username})
```
zyro: Let's define this once rather than in each loop iteration, since it doesn't change between notifications here.

server/core_notification.go
```diff
@@ -0,0 +1,215 @@
...
+	cursorQuery := ""
+	if nc.NotificationID != "" {
+		cursorQuery = " AND id > $3::UUID "
+		params = append(params, nc.NotificationID)
+	}
+
+	rows, err := db.Query(`
+SELECT id, subject, content, code, sender_id, create_time
+FROM notification
+WHERE user_id = $1`+cursorQuery+`
```
zyro: Notifications should be ordered by `create_time`, from newest to oldest I think? Previously expiry time achieved this implicitly, now we should probably order by `create_time` and make that part of the cursor too.

server/core_notification.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		logger.Error("Could not delete notifications.", zap.Error(err))
+		return err
+	}
+
+	return nil
+}
+
+func NotificationSave(logger *zap.Logger, db *sql.DB, notifications map[uuid.UUID][]*api.Notification) error {
+	statements := make([]string, 0, len(notifications))
```
zyro: Not sure this preallocation is accurate, but no way to handle it without measuring each slice in the map I think.

server/core_notification.go
```diff
@@ -0,0 +1,215 @@
...
+	params := make([]interface{}, 0, len(notifications))
```
zyro: Same as above.

server/core_notification.go
```diff
@@ -0,0 +1,215 @@
...
+			params = append(params, userID)
+			params = append(params, un.Subject)
+			params = append(params, un.Content)
+			params = append(params, un.Code)
+			params = append(params, un.CreateTime.Seconds)
+			if un.SenderId != "" {
+				params = append(params, un.SenderId)
+			}
+
+			counter = counter + 8
```
zyro: Is this accurate? Looks to me like the counter should advance by 6 in most cases, or by 7 when there is a sender ID.

server/core_notification.go
mofirouz: I can do ORDER BY create_time ASC - but i don't think it is necessary to embed this in the cursor as the notification ID is deterministic and notification create_time timestamp is always an increasing value.

server/core_notification.go
mofirouz: Yep agreed, best effort preallocation at this point I think.

server/core_notification.go
mofirouz: Good catch.

migrations/sql/20180103142001_initial_schema.sql
```diff
@@ -64,14 +64,14 @@ CREATE TABLE IF NOT EXISTS user_edge (
 );
 
 CREATE TABLE IF NOT EXISTS notification (
-    PRIMARY KEY (id),
-    id              UUID            NOT NULL,
+    PRIMARY KEY (user_id, create_time ASC, id),
```
zyro: I believe `ASC` is the default order. Shall we omit it here?

server/core_notification.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+	}
+
+	params := []interface{}{userID, limit}
+	cursorQuery := " "
+	limitParam := "$2"
+	if nc.NotificationID != nil {
+		cursorQuery = " AND (user_id, create_time, id) > ($1::UUID, $2, $3::UUID)"
+		params = append(params, nc.CreateTime, nc.NotificationID)
+		limitParam = "$4"
```
zyro: The limit param is still `$2` I think, so no need for this special casing.

migrations/sql/20180103142001_initial_schema.sql
novabyte: @zyro `ASC` is left in place for explicitness because the natural order should be `DESC` for notifications but the query analyser with cockroachdb produces really poor query plans when not `ASC` ordered with keyset `WHERE` clauses. It might be useful to leave a note next to it I guess.

novabyte:
@mofirouz Looks good to me 👍

PR #162, opened 2018-02-26
Switch to JSONB and add wallet field.

mofirouz:


novabyte:
@mofirouz Looks good 👍

PR #163, opened 2018-02-27
Add realtime match feature

zyro:


zyro:
@mofirouz Maybe once realtime chat is wired in, but that entity is also used in message history listings which are not done on the socket.

PR #166, opened 2018-03-02
Make proto uniform. Add storage proto and schema

mofirouz:


migrate/sql/20180103142001_initial_schema.sql
```diff
@@ -75,7 +75,29 @@ CREATE TABLE IF NOT EXISTS notification (
     create_time     BIGINT          CHECK (create_time > 0) NOT NULL
 );
 
+CREATE TABLE IF NOT EXISTS storage (
+    PRIMARY KEY (collection, read, key, user_id),
```
zyro: I think `read` shouldn't be part of the primary key. We don't want to let two records with the same collection/key/user_id but different read permissions count as different records.

novabyte:
@mofirouz Looks good 👍

PR #176, opened 2018-03-19
Support direct use of SSL on the HTTP auth/websocket

sinbad:
Presently Nakama requires the use of a separate SSL terminator such as a load balancer. This adds additional complexity, resource usage and cost - if you set up Load Balancer on Google Cloud Platform to do the job, it actually costs more than a small instance to run. You can set up Nginx locally to do it, but this is another moving part which needs configuring & maintaining, uses system resources, and which can potentially fail.

Small devs using Nakama in limited scenarios may want to keep things small and simple, taking the risk of running a single node cheaply (they couldn't run multiple nodes without Enterprise anyway) while still protecting user data by using SSL (you can use self-signed certificates securely, see related PR for Unity client https://github.com/heroiclabs/nakama-unity/pull/60).

This PR allows you to configure an ssl cert/key combo for the Nakama server itself, meaning you can use the server directly instead of via an intermediary. This is clearly not as robust as using 3rd party proxy services to help with DDoS protection etc, but a dev running the server on a single node is already making the trade-off of resilience for simplicity / cost, so it's a valid usage scenario for some.

I'm prepared that this PR might not be accepted since it's not a recommended configuration, but @mofirouz encouraged me to submit it in case it's of use to someone anyway.

### DIFF
--- a/server/session_auth.go
+++ b/server/session_auth.go
@@ -16,6 +16,7 @@ package server
 
 import (
 	"bytes"
+	"crypto/tls"
 	"database/sql"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


server/session_auth.go
```diff
@@ -308,6 +309,17 @@ func (a *authenticationService) configure() {
 	handlerWithCORS := handlers.CORS(CORSHeaders, CORSOrigins)(a.mux)
 
 	a.httpServer = &http.Server{Addr: fmt.Sprintf(":%d", a.config.GetSocket().Port), Handler: handlerWithCORS}
+
+	sockConfig := a.config.GetSocket()
+	if len(sockConfig.SSLCertificate) > 0 && len(sockConfig.SSLPrivateKey) > 0 {
+		cer, err := tls.LoadX509KeyPair(sockConfig.SSLCertificate, sockConfig.SSLPrivateKey)
+		if err != nil {
+			a.logger.Error("Loading SSL certs failed", zap.Error(err))
```
mofirouz: We'll need to make this into FATAL to ensure that the server doesn't start with invalid config values, aka fail fast, as this could make developers confused as to why SSL certs were given but unused. I'll make this code change on master.

sinbad: Ah OK, thanks. I wasn’t sure what the expected error level was and erred on the side of not stopping the startup but that makes sense.

PR #202, opened 2018-05-13
Fix a issue which skips password validation

oott123:
It used to compare `hashedPassword` with `password`, but `hashedPassword` has just generated from `password` !

I fixed it using password fetched from the database.


zyro:
@oott123 Good catch!

PR #207, opened 2018-06-05
Finer control over runtime pooling behaviour.

zyro:
- New runtime pool min/max size options.
- Runtime modules are now loaded in a consistent order.


novabyte:
@zyro Looks good :thumbsup:

PR #209, opened 2018-06-09
Pass through query params for RPC operations.

zyro:
- Expose query params for RPC requests received through the HTTP gateway.
- Users ban & unban functions in script runtime.
- Debug-level logging for WebSocket ping/pong.


server/session_ws.go
```diff
@@ -117,6 +117,7 @@ func (s *sessionWS) Consume(processRequest func(logger *zap.Logger, session Sess
 	s.conn.SetReadDeadline(time.Now().Add(time.Duration(s.config.GetSocket().PongWaitMs) * time.Millisecond))
 	s.conn.SetPongHandler(func(string) error {
 		s.conn.SetReadDeadline(time.Now().Add(time.Duration(s.config.GetSocket().PongWaitMs) * time.Millisecond))
+		s.logger.Debug("Received WebSocket PONG")
```
mofirouz: Do we need this still? I vote we remove it as it effectively makes the debug logs very very verbose.

PR #210, opened 2018-06-12
Mark authoritative match messages with a receive timestamp.

zyro:


main.go
```diff
@@ -152,6 +153,17 @@ func main() {
 		ga.SendSessionStop(http.DefaultClient, gacode, cookie)
 	}
 
+	// TODO
+	fm, err := os.Create("mem.prof")
```
mofirouz: Is this left over from local testing?

PR #213, opened 2018-06-23
Add config for authoritative match buffers.

zyro:


server/config.go
```diff
@@ -41,6 +41,7 @@ type Config interface {
 	GetDatabase() *DatabaseConfig
 	GetSocial() *SocialConfig
 	GetRuntime() *RuntimeConfig
+	GetMatch() *MatchConfig
```
mofirouz: wdyt of calling this config flag `multiplayer`?

zyro: Debated and decided to stick with `match`.

PR #226, opened 2018-08-04
Thread through client addr to Lua context

mofirouz:
_Do not merge - this is untested_

This was a bit of nightmare trying to figure out which info is passed into the correct place to look up the client address info - especially for gRPC.

Turns out that if you only use `peer.FromContext` - it will give you gRPC server info which is masked to localhost if you use grpcGateway. So we need to look up both context metadata headers as well as peerInfo in case the client is HTTP/2. 

Moreover, GCP Loadbalancer has specific rules about X-Forwarded-For header. It appends the IP address in a certain manner. Long story short is that the first element of the X-Forwarded-For header is the real-ip as detected from GCP LB. 

Relevant bits of info:

https://github.com/grpc-ecosystem/grpc-gateway/pull/155/commits/17b6bbc279a5ebcb383abea8ace67f9ad6cf1331

https://cloud.google.com/load-balancing/docs/https/#components

https://husobee.github.io/golang/ip-address/2015/12/17/remote-ip-go.html

### DIFF
--- a/server/runtime_lua_context.go
+++ b/server/runtime_lua_context.go
@@ -58,8 +58,8 @@ const (
 	__CTX_USERNAME         = "username"
 	__CTX_USER_SESSION_EXP = "user_session_exp"
 	__CTX_SESSION_ID       = "session_id"
-	__CTX_IP_ADDRESS       = "ip_address"
-	__CTX_PORT             = "port"
+	__CTX_CLIENT_IP        = "client_ip"
+	__CTX_CLIENT_PORT      = "client_port"
 	__CTX_MATCH_ID         = "match_id"
 	__CTX_MATCH_NODE       = "match_node"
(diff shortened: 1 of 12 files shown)


server/runtime_lua_context.go
```diff
@@ -58,21 +58,31 @@ const (
 	__CTX_USERNAME         = "username"
 	__CTX_USER_SESSION_EXP = "user_session_exp"
 	__CTX_SESSION_ID       = "session_id"
+	__CTX_IP_ADDRESS       = "ip_address"
```
zyro: Maybe use `client_ip` as the key?

server/runtime_lua_context.go
```diff
@@ -58,21 +58,31 @@ const (
...
+	__CTX_PORT             = "port"
```
zyro: Maybe use `client_port` as the key?

PR #246, opened 2018-10-05
Graceful server shutdown and match termination.

zyro:


server/config.go
```diff
@@ -289,14 +299,26 @@ func (c *config) GetLeaderboard() *LeaderboardConfig {
...
 
+// ShutdownConfig is configuration relevant to server shutdown behaviour.
+type ShutdownConfig struct {
+	GracePeriodSec int `yaml:"grace_period_sec" json:"grace_period_sec" usage:"Maximum number of seconds to wait for the server to complete work before shutting down. Default is 60 seconds. If 0 the server will shut down immediately when it receives a termination singal."`
+}
+
+// NewShutdownConfig creates a new ShutdownConfig struct.
+func NewShutdownConfig() *ShutdownConfig {
+	return &ShutdownConfig{
+		GracePeriodSec: 60,
```
novabyte: I wonder whether we should keep it as 0 for backwards compatibility?

mofirouz: I vote for `0` too.

mofirouz:
I'd vote that we name this `--server.shutdown_grace_sec` - this way we can bundle up future generic server configs in the `--server` cmdflag.

mofirouz:
Btw, for what its worth, k8s calls it `terminationGracePeriodSeconds`

PR #247, opened 2018-10-07
Add RELEASE instructions and Docker files.

novabyte:
This is preparation for the Nakama 2.1.0 release which requires CGO_ENABLED for the dynamic linker so that Go plugin code can be loaded by the game server at runtime.

The plugin builder Docker image is a utility to develop plugin code which uses our official Docker images.


build/Dockerfile
```diff
@@ -1,27 +1,42 @@
...
+ENV GOOS linux
+ENV GOARCH amd64
+ENV CGO_ENABLED 1
 
 RUN apk --no-cache add ca-certificates gcc musl-dev git
 
 WORKDIR /go/src/github.com/heroiclabs/
 RUN git config --global advice.detachedHead false && \
-  git clone -q -n https://github.com/heroiclabs/nakama
+  git clone -q -n https://github.com/heroiclabs/nakama nakama
```
mofirouz: can we just clone into `/go/src/github.com/heroiclabs/nakama` directly? This way we can just remove the above WORKDIR line entirely?

build/Dockerfile
```diff
@@ -1,27 +1,42 @@
...
-  GOOS=linux GOARCH=amd64 go build && \
-  mkdir -p /go/build && \
-  mv nakama /go/build
+  go build -o /go/build/nakama -gcflags "-trimpath $PWD" -asmflags "-trimpath $PWD" -ldflags "-s -w -X main.version=$version -X main.commitID=$commit"
 
 FROM alpine:3.7
 
 MAINTAINER Heroic Labs <support@heroiclabs.com>
 
-ARG version
```
mofirouz: I think we might still need this here for docker to pass the `version` arg down to this container?

build/Dockerfile.pluginbuilder
```diff
@@ -0,0 +1,37 @@
+## Copyright 2018 The Nakama Authors
```
mofirouz: I'd suggest calling this `pluginbuilder.Dockerfile` as editors can highlight the file properly.

As a better alternative, and what the community convention is, create a folder called `pluginbuilder` and put this file as simply `Dockerfile` there.

mofirouz: https://github.com/docker-library/golang/tree/ed78459fac108dab72556146b759516cc65ee109/1.11 or https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles

novabyte:
@mofirouz I've made the requested changes. Please have another look and let me know it's good to merge.

PR #251, opened 2018-10-18
Fix racy timers for websocket pinging

zengjie:
According to [Go docs](https://golang.org/pkg/time/#Timer.Reset), it should not be done concurrent to other receives from the Timer's channel. This will cause panics for high server loads.

Additionally, it's better to use defer func to ensure Unlock() is called before exiting pingNow(), IMHO.

### DIFF
--- a/server/session_ws.go
+++ b/server/session_ws.go
@@ -65,6 +65,7 @@ type sessionWS struct {
 	pingTimer              *time.Timer
 	outgoingCh             chan []byte
 	outgoingStopCh         chan struct{}
+	pingTimerResetCh       chan time.Duration
 }
 
 func NewSessionWS(logger *zap.Logger, config Config, userID uuid.UUID, username string, expiry int64, clientIP string, clientPort string, jsonpbMarshaler *jsonpb.Marshaler, jsonpbUnmarshaler *jsonpb.Unmarshaler, conn *websocket.Conn, sessionRegistry *SessionRegistry, matchmaker Matchmaker, tracker Tracker) Session {
@@ -105,6 +106,7 @@ func NewSessionWS(logger *zap.Logger, config Config, userID uuid.UUID, username 
 		pingTimer:              time.NewTimer(time.Duration(config.GetSocket().PingPeriodMs) * time.Millisecond),
 		outgoingCh:             make(chan []byte, config.GetSocket().OutgoingQueueSize),
 		outgoingStopCh:         make(chan struct{}),
+		pingTimerResetCh:       make(chan time.Duration),
 	}
 }
 
@@ -149,7 +151,7 @@ func (s *sessionWS) Consume(processRequest func(logger *zap.Logger, session Sess
 	s.conn.SetReadLimit(s.config.GetSocket().MaxMessageSizeBytes)
 	s.conn.SetReadDeadline(time.Now().Add(s.pongWaitDuration))
 	s.conn.SetPongHandler(func(string) error {
-		s.pingTimer.Reset(s.pingPeriodDuration)
+		s.pingTimerResetCh <- s.pingPeriodDuration
 		s.conn.SetReadDeadline(time.Now().Add(s.pongWaitDuration))
 		return nil
 	})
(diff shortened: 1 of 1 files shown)


zengjie:
I also changed logging level for "Error reading message from client" to make us always know why the client connection is closed.

server/session_ws.go
```diff
@@ -164,7 +166,7 @@ func (s *sessionWS) Consume(processRequest func(logger *zap.Logger, session Sess
 			if !websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway, websocket.CloseNoStatusReceived) {
 				// Ignore underlying connection being shut down while read is waiting for data.
 				if e, ok := err.(*net.OpError); !ok || e.Err.Error() != "use of closed network connection" {
-					s.logger.Debug("Error reading message from client", zap.Error(err))
+					s.logger.Warn("Error reading message from client", zap.Error(err))
```
zyro: At high concurrent user counts there will always be some number of disconnects expected, usually due to end device network connectivity problems. This is logged at debug level because it's not a problem for the server or something for the developer/operator to fix, and will cause a large amount of log noise.

I suggest we leave this at debug level.

server/session_ws.go
```diff
@@ -233,13 +237,13 @@ func (s *sessionWS) processOutgoing() {
 
 func (s *sessionWS) pingNow() bool {
 	s.Lock()
+	defer s.Unlock()
```
zyro: The lock is handled explicitly in a few situations so we can unlock as early as possible, before the function returns. With this change the function will also deadlock the session when it exits due to an error because it calls `s.cleanupClosedConnection()` which tries to acquire its own lock.

I suggest we do not change the locking strategy here to avoid a cascade of further changes needed to make it work.

server/session_ws.go
```diff
@@ -173,7 +175,7 @@ func (s *sessionWS) Consume(processRequest func(logger *zap.Logger, session Sess
 		s.receivedMessageCounter--
 		if s.receivedMessageCounter <= 0 {
 			s.receivedMessageCounter = s.config.GetSocket().PingBackoffThreshold
-			s.pingTimer.Reset(s.pingPeriodDuration)
+			s.pingTimerResetCh <- s.pingPeriodDuration
```
zyro: The ping backoff exists to avoid unnecessary context switches caused by session ping/pong at high concurrent user counts when the socket is known to be constantly active. Introducing a channel here might bring back this impact, but I definitely see the case where concurrent access is possible and should be prevented.

What do you think of using https://github.com/heroiclabs/nakama/commit/ac9810151dccc9ef4db8e2b2d893c08158996caf to achieve this instead?

server/session_ws.go
zengjie: OK. I think you are right. If the network of clients are bad, it will log a lot.

server/session_ws.go
zengjie: I agree. I made this change because pingTimer.Reset is not protected by the lock. If there's a better way to do this, there's no need to change the locking strategy.

server/session_ws.go
zyro: I see your point, I believe https://github.com/heroiclabs/nakama/commit/7a3a800d2fcfb34360712189e3307bc09358b159 handles it correctly then. Thoughts?

zengjie: Awesome, it's correct now. I've just read [time.go](https://golang.org/src/runtime/time.go) source code and the problematic session_ws.go. The root cause of the panics was timer.Reset is called in both receiving goroutine and sending goroutine and will cause a racy access [as this gist shown](https://gist.github.com/zengjie/ec6194c5f8d2b6eefc2e2a8340c29e7f). Now, we have only resets in receiving goroutine, there will be no panic.

However, I am wondering how to maintain ping period if we don't reset the timer in pingNow. I think it's ok to reset in pingNow with the proposed "stop and check" protection.

zyro: The timer will be reset when the corresponding pong is received. If the pong never makes it then we're closing the session anyway and there's no point in resetting the timer.

I think we have our solution then!

zyro:
Solved in https://github.com/heroiclabs/nakama/commit/e48928bea9609e73127aa6d6c3fac9287b5a3d20.

zyro:
@zengjie Thanks for the help on this!

PR #252, opened 2018-10-19
Add type info to wallet update error message

Grimeh:
Creating a `nk.WalletUpdate` request like:
```
	var changeset = map[string]interface{}{
		"coins": 100,
	}
	err := nk.WalletUpdate(ctx, userID, changeset, nil)
	if err != nil {
		logger.Printf("failed to update user wallet: %v", err)
		return "", errors.New("unable to update wallet")
	}
```
results in a very confusing & unhelpful error message:
> failed to update user wallet: unknown update changeset value type at path 'coins'

Afaik there's no documentation to find what types are expected, the only source of that information is grep-ing through Nakama source. I've added the expected types to the error message to clear up this confusion.

Does this seem reasonable? Is there some documentation that users are expected to have read before reaching this stage that would have explained the types that are expected in this request (that I may have overlooked)?

### DIFF
--- a/server/core_wallet.go
+++ b/server/core_wallet.go
@@ -247,7 +247,7 @@ func applyWalletUpdate(wallet map[string]interface{}, changeset map[string]inter
 				}
 			} else {
 				// Existing value is not a map or float.
-				return nil, errors.Errorf("unknown existing wallet value type at path '%v'", currentPath)
+				return nil, errors.Errorf("unknown existing wallet value type at path '%v'. Expecting map or float64", currentPath)
 			}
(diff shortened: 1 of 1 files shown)


novabyte:
Hi @Grimeh. Thanks for the pull request!
The documentation which covers the virtual wallet is [here](https://heroiclabs.com/docs/user-accounts/#virtual-wallet). Unfortunately it does not cover the constraints of the API. I'll make a note to add more detail to the docs.

zyro:
@Grimeh Thanks!

PR #256, opened 2018-10-22
Fixed a offset-by-one bug in leaderboard record deleting

zengjie:
For example, the old ranks are A:1, B:2, C:3, therefore totalRanks is 3.
When A is deleted, the remained are B:2, C:3 and the shifting should be from index 0 to index 1 which is (totalRanks-1) -1 instead of totalRanks-1.
When B is deleted, the remained are A:1, C:3 and the shifting should be from index 1 to index 1 which is (totalRanks-1) -1 instead of totalRanks-1.

### DIFF
--- a/server/leaderboard_rank_cache.go
+++ b/server/leaderboard_rank_cache.go
@@ -16,10 +16,11 @@ package server
 
 import (
 	"database/sql"
-	"github.com/heroiclabs/nakama/api"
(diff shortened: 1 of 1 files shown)


zyro:
Duplicate of https://github.com/heroiclabs/nakama/pull/255, closing without merge.

PR #260, opened 2018-11-06
Allow sending arbitrary messages instead of StreamData to stream in Go runtime

zengjie:
It's a small step to the vision described in [the Streams doc](https://heroiclabs.com/docs/advanced-streams/#built-in-streams). Sometimes, we need to change the behaviors of built-in streams.


zyro:
@zengjie Thanks!

PR #280, opened 2018-12-21
Add new Go runtime logger interface.

zyro:


server/api_authenticate.go
```diff
@@ -217,37 +217,60 @@ func (s *ApiServer) AuthenticateEmail(ctx context.Context, in *api.AuthenticateE
...
-		return nil, status.Error(codes.InvalidArgument, "Password must be longer than 8 characters.")
 	} else if !emailRegex.MatchString(email.Email) {
 		return nil, status.Error(codes.InvalidArgument, "Invalid email address format.")
 	} else if len(email.Email) < 10 || len(email.Email) > 255 {
 		return nil, status.Error(codes.InvalidArgument, "Invalid email address, must be 10-255 bytes.")
 	}
 
-	cleanEmail := strings.ToLower(email.Email)
+	if len(email.Password) < 8 {
+		return nil, status.Error(codes.InvalidArgument, "Password must be longer than 8 characters.")
```
Dimon4eg: Must be 8 or longer

server/core_authenticate.go
```diff
@@ -275,6 +276,44 @@ func AuthenticateEmail(ctx context.Context, logger *zap.Logger, db *sql.DB, emai
...
+		} else {
+			logger.Error("Error looking up user by username.", zap.Error(err), zap.String("username", username))
+			return "", status.Error(codes.Internal, "Error finding user account.")
+		}
+	}
+
+	// Check if it's disabled.
+	if dbDisableTime.Valid && dbDisableTime.Time.Unix() != 0 {
+		logger.Info("User account is disabled.", zap.String("username", username))
+		return "", status.Error(codes.Unauthenticated, "Error finding or creating user account.")
```
Dimon4eg: Maybe better to return more clear error description, the same as logged?

zyro: This is intentionally vague. The developer gets a descriptive log message but under the current API contract clients are not told when their account is disabled.

server/api_authenticate.go
zyro: Unfortunately the current API error response structure means clients rely on error strings to disambiguate between various bad input cases. I don't think we should change it in case it breaks client error handling

novabyte:
@zyro Do you think we should add functions like `IsDebugEnabled` to the new logger interface?

zyro:
As discussed the level check done in each logger function is enough for now.

PR #287, opened 2019-01-06
Fix for unmarhalling steam id

nielslanting:
```
{"level":"info","ts":"2019-01-07T00:15:23.637+0100","msg":"Could not authenticate Steam profile.","error":"json: cannot unmarshal string into Go struct field SteamProfile.steamid of type uint64"}
```


zyro:
@nielslanting Awesome, thanks! ⭐️

PR #301, opened 2019-02-26
google-profile-update-vg

gupta-vaibhav:
The existing code pulls the profile details, i.e. DisplayName and AvatarUrl from the googleProfile but doesn't save it in the db. Thus the player's updated profile is not reflected in the game. This pull request fixes that.

Please note the following:

1. Changes have been made to api_link (when a Google Account is linked to an existing account) and core_authenticate (when a new Google Account is created, or changes are made to an existing Google Account)
2. The error in updating DisplayName and AvatarUrl have been suppressed, since this is not an error which should interrupt the code execution. The update will be retried automatically at the time of the next login.

Notes for testing:
At times changes to Google Profile take time to reflect, thus the changes to the Google Account may not show up immediately.

### DIFF
--- a/server/api_link.go
+++ b/server/api_link.go
@@ -15,6 +15,9 @@
 package server
 
 import (
+	"strconv"
+	"strings"
+
 	"github.com/cockroachdb/cockroach-go/crdb"
 	"github.com/gofrs/uuid"
 	"github.com/golang/protobuf/ptypes/empty"
@@ -25,8 +28,6 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"strconv"
-	"strings"
 )
 
 func (s *ApiServer) LinkCustom(ctx context.Context, in *api.AccountCustom) (*empty.Empty, error) {
@@ -441,16 +442,18 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
 		s.logger.Info("Could not authenticate Google profile.", zap.Error(err))
 		return nil, status.Error(codes.Unauthenticated, "Could not authenticate Google profile.")
 	}
-	
+
 	displayName := googleProfile.Name
 	if len(displayName) > 255 {
-		// Trim the name in case it is longer than db can store
-		displayName = displayName[0:255]		
+		// Ignore the name in case it is longer than db can store
+		s.logger.Warn("Skipping updating display_name: value received from Google longer than max length of 255 chars.", zap.String("display_name", displayName))
+		displayName = ""
 	}
-	
+
 	avatarUrl := googleProfile.Picture
-	if len(avatarUrl) > 512 {
+	if len(avatarUrl) > 512 || avatarUrl == "" {
 		// Ignore the url in case it is longer than db can store
+		s.logger.Warn("Skipping updating avatar_url: value received from Google longer than max length of 512 chars.", zap.String("avatar_url", avatarUrl))
 		avatarUrl = ""
 	}
 
--- a/server/api_link.go
+++ b/server/api_link.go
@@ -444,14 +444,14 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
 
 	res, err := s.db.ExecContext(ctx, `
 UPDATE users
-SET google_id = $2, update_time = now()
+SET google_id = $2, display_name = $3, avatar_url = $4, update_time = now()
 WHERE (id = $1)
 AND (NOT EXISTS
     (SELECT id
      FROM users
      WHERE google_id = $2 AND NOT id = $1))`,
 		userID,
-		googleProfile.Sub)
+		googleProfile.Sub, googleProfile.Name, googleProfile.Picture)
 
 	if err != nil {
 		s.logger.Error("Could not link Google ID.", zap.Error(err), zap.Any("input", in))
--- a/server/api_link.go
+++ b/server/api_link.go
@@ -441,6 +441,18 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
 		s.logger.Info("Could not authenticate Google profile.", zap.Error(err))
 		return nil, status.Error(codes.Unauthenticated, "Could not authenticate Google profile.")
 	}
+	
+	displayName := googleProfile.Name
+	if len(displayName) > 255 {
+		// Trim the name in case it is longer than db can store
+		displayName = displayName[0:255]		
+	}
+	
+	avatarUrl := googleProfile.Picture
+	if len(avatarUrl) > 512 {
+		// Ignore the url in case it is longer than db can store
+		avatarUrl = ""
+	}
 
 	res, err := s.db.ExecContext(ctx, `
 UPDATE users
@@ -451,7 +463,7 @@ AND (NOT EXISTS
      FROM users
      WHERE google_id = $2 AND NOT id = $1))`,
 		userID,
-		googleProfile.Sub, googleProfile.Name, googleProfile.Picture)
+		googleProfile.Sub, displayName, avatarUrl)
 
 	if err != nil {
 		s.logger.Error("Could not link Google ID.", zap.Error(err), zap.Any("input", in))
--- a/server/core_authenticate.go
+++ b/server/core_authenticate.go
@@ -104,6 +104,7 @@ func AuthenticateDevice(ctx context.Context, logger *zap.Logger, db *sql.DB, dev
 
 	// Look for an existing account.
 	query := "SELECT user_id FROM user_device WHERE id = $1"
+	logger.Info(query, zap.String("userId", deviceID))
 	var dbUserID string
 	err := db.QueryRowContext(ctx, query, deviceID).Scan(&dbUserID)
 	if err != nil {
@@ -471,6 +472,18 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
 		}
 	}
 
+	displayName := googleProfile.Name
+	if len(displayName) > 255 {
+		// Trim the name in case it is longer than db can store
+		displayName = displayName[0:255]		
+	}
+	
+	avatarUrl := googleProfile.Picture
+	if len(avatarUrl) > 512 {
+		// Ignore the url in case it is longer than db can store
+		avatarUrl = ""
+	}
+
 	// Existing account found.
 	if found {
 		// Check if it's disabled.
@@ -480,14 +493,12 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
 		}
 
 		// Check if display name / avatar received from Google different from those in the DB
-		if dbDisplayName.String != googleProfile.Name || dbAvatarUrl.String != googleProfile.Picture {
-			logger.Info("About to check Google Id's details now. Found differences")
+		if dbDisplayName.String != displayName || dbAvatarUrl.String != avatarUrl {
 			// At least one difference found, update the DB to reflect changed values
-			if _, err = db.ExecContext(ctx, "UPDATE users SET display_name = $1, avatar_url = $2, update_time = now() WHERE google_id = $3", googleProfile.Name, googleProfile.Picture, googleProfile.Sub); err != nil {
-				logger.Error("Error in updating google profile details", zap.Error(err), zap.String("googleId", googleProfile.Sub), zap.String("displayName", googleProfile.Name), zap.String("avatarUrl", googleProfile.Picture))
+			if _, err = db.ExecContext(ctx, "UPDATE users SET display_name = $1, avatar_url = $2, update_time = now() WHERE id = $3", displayName, avatarUrl, dbUserID); err != nil {
+				logger.Error("Error in updating google profile details", zap.Error(err), zap.String("googleId", googleProfile.Sub), zap.String("display_name", googleProfile.Name), zap.String("display_name", googleProfile.Picture))
 				// Do no return error. Failure to update the display_name / avatar_url should not
 				// interrupt the execution. Ignore the result and the error, just log the error.
-				// return "", "", false, status.Error(codes.Unknown, "Error finding or creating user account.")
 			}
 		}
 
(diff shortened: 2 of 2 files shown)


server/core_authenticate.go
```diff
@@ -477,6 +479,18 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
 			return "", "", false, status.Error(codes.Unauthenticated, "Error finding or creating user account.")
 		}
 
+		// Check if display name / avatar received from Google different from those in the DB
+		if dbDisplayName.String != googleProfile.Name || dbAvatarUrl.String != googleProfile.Picture {
+			logger.Info("About to check Google Id's details now. Found differences")
```
mofirouz: I think this log line should be removed.

server/core_authenticate.go
```diff
@@ -477,6 +479,18 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
...
+			// At least one difference found, update the DB to reflect changed values
+			if _, err = db.ExecContext(ctx, "UPDATE users SET display_name = $1, avatar_url = $2, update_time = now() WHERE google_id = $3", googleProfile.Name, googleProfile.Picture, googleProfile.Sub); err != nil {
```
mofirouz: We should try and hit the primary index if we can - so I'd recommend changing the query to read:

```
UPDATE users SET display_name = $1, avatar_url = $2, update_time = now() WHERE id = $3
```

Since you already have the userID from a previous query.

server/core_authenticate.go
```diff
@@ -477,6 +479,18 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
...
+				logger.Error("Error in updating google profile details", zap.Error(err), zap.String("googleId", googleProfile.Sub), zap.String("displayName", googleProfile.Name), zap.String("avatarUrl", googleProfile.Picture))
```
mofirouz: The logger uses the JSON format. Please change `displayName` to `display_name` and `avatarUrll` to `avatar_url` to adhere to the JSON naming conventions.

server/core_authenticate.go
```diff
@@ -477,6 +479,18 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
...
+				// Do no return error. Failure to update the display_name / avatar_url should not
+				// interrupt the execution. Ignore the result and the error, just log the error.
+				// return "", "", false, status.Error(codes.Unknown, "Error finding or creating user account.")
```
mofirouz: We can remove this line:

```
// return "", "", false, status.Error(codes.Unknown, "Error finding or creating user account.")
```

Your text above is clear :)

server/api_link.go
```diff
@@ -444,14 +444,14 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
...
-SET google_id = $2, update_time = now()
+SET google_id = $2, display_name = $3, avatar_url = $4, update_time = now()
 WHERE (id = $1)
 AND (NOT EXISTS
     (SELECT id
      FROM users
      WHERE google_id = $2 AND NOT id = $1))`,
 		userID,
-		googleProfile.Sub)
+		googleProfile.Sub, googleProfile.Name, googleProfile.Picture)
```
mofirouz: What's the format of Name and Picture? Does it work within the constraints of the columns in the database?

https://github.com/heroiclabs/nakama/blob/master/migrate/sql/20180103142001_initial_schema.sql#L23

server/core_authenticate.go
gupta-vaibhav: Yes! Removed in the next commit.

server/core_authenticate.go
gupta-vaibhav: Spot on! Updated in the next commit.

server/core_authenticate.go
gupta-vaibhav: Changed in the next commit, although I had followed the format used for 'googleId' in the code elsewhere.

server/core_authenticate.go
gupta-vaibhav: Removed in the next commit.

server/api_link.go
gupta-vaibhav: Nice observation. Both the fields are string, however length of the text could be a corner case that needs to be handled. Have revised the code to the following, which you will see in the next commit.

Created and used the following temp variables instead of the values received in googleProfile.
```
	displayName := googleProfile.Name
	if len(displayName) > 255 {
		// Trim the name in case it is longer than db can store
		displayName = displayName[0:255]		
	}
	
	avatarUrl := googleProfile.Picture
	if len(avatarUrl) > 512 {
		// Ignore the url in case it is longer than db can store
		avatarUrl = ""
	}
```

gupta-vaibhav:
@novabyte have pushed another commit which should handle the feedback and questions raised by you, I hope. Do let me know in case there's more that needs to be discussed.

server/api_link.go
```diff
@@ -441,17 +441,29 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
 		s.logger.Info("Could not authenticate Google profile.", zap.Error(err))
 		return nil, status.Error(codes.Unauthenticated, "Could not authenticate Google profile.")
 	}
+	
+	displayName := googleProfile.Name
+	if len(displayName) > 255 {
+		// Trim the name in case it is longer than db can store
+		displayName = displayName[0:255]
```
mofirouz: I think we should let displayName to be "" if we can't fully update it. We should also leave a WARN log to tell developers that this has happened.

server/api_link.go
```diff
@@ -441,17 +441,29 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
...
+	displayName := googleProfile.Name
+	if len(displayName) > 255 {
+		// Trim the name in case it is longer than db can store
+		displayName = displayName[0:255]		
+	}
+	
+	avatarUrl := googleProfile.Picture
+	if len(avatarUrl) > 512 {
+		// Ignore the url in case it is longer than db can store
+		avatarUrl = ""
```
mofirouz: We should also leave a WARN log to tell developers that this has happened.

server/core_authenticate.go
```diff
@@ -104,6 +104,7 @@ func AuthenticateDevice(ctx context.Context, logger *zap.Logger, db *sql.DB, dev
 
 	// Look for an existing account.
 	query := "SELECT user_id FROM user_device WHERE id = $1"
+	logger.Info(query, zap.String("userId", deviceID))
```
mofirouz: Let's remove this log line.

server/core_authenticate.go
```diff
@@ -469,6 +472,18 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
 		}
 	}
 
+	displayName := googleProfile.Name
+	if len(displayName) > 255 {
+		// Trim the name in case it is longer than db can store
+		displayName = displayName[0:255]
```
mofirouz: I think we should let displayName to be "" if we can't fully update it. We should also leave a WARN log to tell developers that this has happened.

server/core_authenticate.go
```diff
@@ -469,6 +472,18 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
...
+	displayName := googleProfile.Name
+	if len(displayName) > 255 {
+		// Trim the name in case it is longer than db can store
+		displayName = displayName[0:255]		
+	}
+	
+	avatarUrl := googleProfile.Picture
+	if len(avatarUrl) > 512 {
+		// Ignore the url in case it is longer than db can store
+		avatarUrl = ""
```
mofirouz: We should also leave a WARN log to tell developers that this has happened.

server/core_authenticate.go
```diff
@@ -477,6 +492,16 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
 			return "", "", false, status.Error(codes.Unauthenticated, "Error finding or creating user account.")
 		}
 
+		// Check if display name / avatar received from Google different from those in the DB
+		if dbDisplayName.String != displayName || dbAvatarUrl.String != avatarUrl {
```
mofirouz: Note for the team: We are now introducing a hit to the database on re-authentication (which we've never had before). I'm OK with this exception because I assume this will be an infrequent operation (i.e. users won't update their name/avatar frequently).

@novabyte @zyro Post your thoughts here, and if OK we can merge this.

server/api_link.go
gupta-vaibhav: Well! I believe we should truncate the name, or another option could be:
```
	displayName = displayName[0:252] + "..."
```
since (almost) no game will ever display a name that long in full. In most games that I've seen / worked on, we already start truncating the names longer than 20 chars.

server/api_link.go
gupta-vaibhav: Agreed, will add.

server/core_authenticate.go
gupta-vaibhav: Will remove.

server/core_authenticate.go
gupta-vaibhav: On the same boat as @mofirouz. Evaluating the situation in totality, downside of avoiding this database hit is showing an outdated name / profile pic to the user which would be worse, IMO.

server/api_link.go
mofirouz: I side with accuracy, i.e. if we can't update probably it shouldn't be updated. @zyro @novabyte thoughts please.

server/core_authenticate.go
zyro: I'm OK with this tradeoff. Let's try and keep the info in sync - it should not change often enough to have an impact.

server/api_link.go
zyro: I vote for correctness. If the display name is too long, ignore it and log a warning.

server/api_link.go
```diff
@@ -441,17 +441,29 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
...
+	}
 
 	res, err := s.db.ExecContext(ctx, `
 UPDATE users
-SET google_id = $2, update_time = now()
+SET google_id = $2, display_name = $3, avatar_url = $4, update_time = now()
```
zyro: This query should change. Right now if the profile did not have a display name or avatar URL (or they were too long) the query is wiping out any existing values. If display name or avatar URL are `""` the query should not `SET` them.

zyro:
A few small comments left, otherwise a really nice change @gupta-vaibhav!

server/api_link.go
gupta-vaibhav: Super, will change it to be that way.

server/api_link.go
gupta-vaibhav: @zyro @mofirouz while I agree that if the display name is too long it should not wipe out the existing one. But this can be done in two ways as follows:

**Option 1**: Reset the values only if new value > max length
```
	displayName := googleProfile.Name
	if len(displayName) > 255 {
		// Trim the name in case it is longer than db can store
		logger.Warn("Skipping updating display_name: value received from Google longer than max length of 255 chars.", zap.String("display_name", displayName))
		displayName = dbDisplayName.String
	}
	
	avatarUrl := googleProfile.Picture
	if len(avatarUrl) > 512 {
		// Ignore the url in case it is longer than db can store
		logger.Warn("Skipping updating avatar_url: value received from Google longer than max length of 512 chars.", zap.String("avatar_url", avatarUrl))
		avatarUrl = dbAvatarUrl.String
	}
```

**Option 2**: Reset the values if either new value > max length and also if new value = ''
```
	displayName := googleProfile.Name
	if len(displayName) > 255 {
		// Trim the name in case it is longer than db can store
		logger.Warn("Skipping updating display_name: value received from Google longer than max length of 255 chars.", zap.String("display_name", displayName))
		displayName = ""
	}
	
	avatarUrl := googleProfile.Picture
	if len(avatarUrl) > 512 {
		// Ignore the url in case it is longer than db can store
		logger.Warn("Skipping updating avatar_url: value received from Google longer than max length of 512 chars.", zap.String("avatar_url", avatarUrl))
		avatarUrl = ""
	}

	/* .... other code.... */

	if dbDisplayName.String != displayName || dbAvatarUrl.String != avatarUrl {
		// At least one difference found, update the DB to reflect changed values
		if _, err = db.ExecContext(ctx, "UPDATE users SET "+
			"display_name = CASE WHEN $1 = '' then display_name ELSE $1 END, "+
			"avatar_url = CASE WHEN $2 = '' THEN avatar_url ELSE $2 END, "+
			"update_time = now() WHERE id = $3", displayName, avatarUrl, dbUserID); err != nil {
			logger.Error("Error in updating google profile details", zap.Error(err), zap.String("googleId", googleProfile.Sub), zap.String("display_name", googleProfile.Name), zap.String("display_name", googleProfile.Picture))
			// Do no return error. Failure to update the display_name / avatar_url should not
			// interrupt the execution. Ignore the result and the error, just log the error.
		}
	}
```

I prefer Option 1, since it gives the user an ability to remove display name if they so wish. Let me know your thoughts and I will close it accordingly.

gupta-vaibhav:
@zyro @mofirouz thanks. Please see my comment on handling long and blank display name / avatar url and will push another commit based on your response.

server/api_link.go
gupta-vaibhav: My bad, a better implementation of Option 2 that will avoid unnecessary database query would be to do just the following, and leave the database query as is:

```
	displayName := googleProfile.Name
	if len(displayName) > 255 || displayName == "" {
		// Trim the name in case it is longer than db can store
		logger.Warn("Skipping updating display_name: value received from Google either blank or longer than max length of 255 chars.", zap.String("display_name", displayName))
		displayName = dbDisplayName.String
	}
	
	avatarUrl := googleProfile.Picture
	if len(avatarUrl) > 512 || avatarUrl == "" {
		// Ignore the url in case it is longer than db can store
		logger.Warn("Skipping updating avatar_url: value received from Google either blank or longer than max length of 512 chars.", zap.String("avatar_url", avatarUrl))
		avatarUrl = dbAvatarUrl.String
	}
```

mofirouz: Maybe I am missing something here but where does `dbDisplayName.String` come from? There is no previous read from the db?

gupta-vaibhav:
Hi guys, I have pushed the changes the way I thought makes sense. For your review.

server/api_link.go
```diff
@@ -441,16 +442,18 @@ func (s *ApiServer) LinkGoogle(ctx context.Context, in *api.AccountGoogle) (*emp
...
-		displayName = displayName[0:255]		
+		// Ignore the name in case it is longer than db can store
+		s.logger.Warn("Skipping updating display_name: value received from Google longer than max length of 255 chars.", zap.String("display_name", displayName))
+		displayName = ""
 	}
-	
+
 	avatarUrl := googleProfile.Picture
-	if len(avatarUrl) > 512 {
+	if len(avatarUrl) > 512 || avatarUrl == "" {
```
mofirouz: Nitpicking - Don't think we need the `|| avatarUrl == ""` in the if statement? 

If we do, we need to update the log message to say "value is either empty or longer than max length of 512 chars."

mofirouz:
LGTM - I'll merge this. Thanks a lot @gupta-vaibhav, brilliant work!

PR #303, opened 2019-03-03
Added constants for storage permissions.

volknanebo:
Added storage permissions with new constants.


Dimon4eg:
@mofirouz I think we need these constants in all clients. I can add to C++ client.

mofirouz:
@Dimon4eg Let's do it - but please add it as a PR (on a branch) as I need to speak to the team for more info on it.

zyro:
@volknanebo Thanks!

PR #310, opened 2019-03-14
Console users handler functions.

zyro:


novabyte:
@zyro This all looks good to me. 👍  My only thought on it is whether we should include default values (empty) in the JSON responses or not.

We'll have no problem with it in the devconsole codebase but it might be more straightforward for other API consumers especially as we don't need to worry about bytes over the wire with an internal administrative API. What do you think?

zyro:
Including default values sounds good to me but the gateway doesn't expose this easily. We may need to replace the [default marshaler implementation](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/marshal_jsonpb.go). I'll look into it further.

PR #312, opened 2019-03-20
:bug: extract directory from logger filename in config, then try to c…

ZhangGuangxu:
Should not treat logger filename in config as a directory. First, extract the logger directory. Then, try to create logger directory if it does not exist.


zyro:
@ZhangGuangxu Thanks! We'll schedule this in for the 2.5.0 release.

PR #318, opened 2019-03-28
Update iap_verifier_rpc.lua

grant-sessiongames:
Updated function google_verify_payment to take the is_subscription value from the payload instead of being explicitly set to failse.


novabyte:
Thanks @grant-sessiongames 👍

PR #321, opened 2019-04-05
Console

revnode:
Milestone 2 work.


console/ui/src/store/storage/sagas.ts
```diff
@@ -0,0 +1,295 @@
...
+  storageDeleteSuccess,
+  storageDeleteError
+} from './actions';
+import {NakamaApi} from '../../api.gen';
+
+function* handleFetchMany({payload: data}: AnyAction)
+{
+  try
+  {
+    const nakama = NakamaApi({
```
novabyte: Same as above.

console/ui/src/store/storage/sagas.ts
```diff
@@ -0,0 +1,295 @@
...
+    localStorage.clear();
+    window.location.href = '/login';
+  }
+}
+
+function* handleDeleteMany()
+{
+  try
+  {
+    const nakama = NakamaApi({
```
novabyte: Same as above.

console/ui/src/store/storage/sagas.ts
```diff
@@ -0,0 +1,295 @@
...
+    localStorage.clear();
+    window.location.href = '/login';
+  }
+}
+
+function* handleCreate({payload: data}: AnyAction)
+{
+  try
+  {
+    const nakama = NakamaApi({
```
novabyte: Same as above.

console/ui/src/store/storage/sagas.ts
```diff
@@ -0,0 +1,295 @@
...
+    localStorage.clear();
+    window.location.href = '/login';
+  }
+}
+
+function* handleFetch({payload: data}: AnyAction)
+{
+  try
+  {
+    const nakama = NakamaApi({
```
novabyte: Same as above.

console/ui/src/store/storage/sagas.ts
```diff
@@ -0,0 +1,295 @@
...
+    localStorage.clear();
+    window.location.href = '/login';
+  }
+}
+
+function* handleUpdate({payload: data}: AnyAction)
+{
+  try
+  {
+    const nakama = NakamaApi({
```
novabyte: Same as above.

console/ui/src/store/storage/sagas.ts
```diff
@@ -0,0 +1,295 @@
...
+    localStorage.clear();
+    window.location.href = '/login';
+  }
+}
+
+function* handleDelete({payload: data}: AnyAction)
+{
+  try
+  {
+    const nakama = NakamaApi({
```
novabyte: Same as above.

mofirouz:
In addition to what @novabyte mentioned, I've found the following when running the dashboard via `yarn serve` and the server on port 80, using Safari and Chrome 73.

Notice - You'll need to run the server with the additional flag `--console.max_message_size_bytes 50000000` to allow for large values to be sent to the serve. We'll adjust defaults.

1. When logging in, see the message in Chrome 73: "Unchecked runtime.lastError: The message port closed before a response was received."

2. When updating a storage record fails, dashboard tries to load Storage view with a bad URL format. "Fetch failed loading: GET "http://127.0.0.1/v2/console/storage?"" and it fails, thus kicked to the login screen.

3. Stop refreshing blindly to "/login" or errors. Only kick to login on error code 401, otherwise please leave the page as it is. I think this goes hand in hand with issue 5 below.

4. Timestamp of updated Storage objects do not refresh when the update is done for the first time.

5. The login error is wiped due to a page reload. See attached GIF.
badusername

mofirouz:
6. Bad layout on Configuration page. See image below.
7. Missing values on the configuration page (although they are available in the exported YAML file).
missingconfig

8. Bad layout of warnings. See image (although may be fixed with what @novabyte said above). I think it is best to combine all warnings into one giant yellow box?
toomuchyellow

mofirouz:
9. When filtering storage by user ID, and then updating a given object, the view is punted back to the storage listing view and have forgotten the filtered user ID. Super annoying.

console/ui/src/routes/configuration/index.tsx
```diff
@@ -41,7 +96,12 @@ class Configuration extends Component
               </Field.Label>
               <Field.Body>
                 <Control>
-                  <Input static type="text" placeholder="(empty)" value="nakama-0" />
+                  <Input
```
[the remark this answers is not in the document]
revnode: Yup, done.

console/ui/src/routes/configuration/index.tsx
```diff
@@ -233,15 +378,45 @@ class Configuration extends Component
...
+                    static
+                    type="text"
+                    placeholder="(empty)"
+                    defaultValue={data.config && data.config.socket && data.config.socket.max_message_size_bytes}
+                  />
                 </Control>
               </Field.Body>
             </Field>
+            <br /><br />
+            {
```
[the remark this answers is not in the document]
revnode: Yup, done.

console/ui/src/store/configuration/sagas.ts
```diff
@@ -0,0 +1,51 @@
+import {all, call, fork, put, takeEvery, select} from 'redux-saga/effects';
+import {ConfigurationActionTypes} from './types';
+import {configurationError, configurationSuccess} from './actions';
+import {NakamaApi} from '../../api.gen';
+
+function* handleFetch()
+{
+  try
+  {
+    const nakama = NakamaApi({
```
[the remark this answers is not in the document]
revnode: Yup, done.

revnode:
> When logging in, see the message in Chrome 73: "Unchecked runtime.lastError: The message port closed before a response was received."

Not seeing this error on Chrome. Maybe it got resolved?

> When updating a storage record fails, dashboard tries to load Storage view with a bad URL format. "Fetch failed loading: GET "http://127.0.0.1/v2/console/storage?"" and it fails, thus kicked to the login screen.
> Stop refreshing blindly to "/login" or errors. Only kick to login on error code 401, otherwise please leave the page as it is. I think this goes hand in hand with issue 5 below.

Yup, fixed.

> Timestamp of updated Storage objects do not refresh when the update is done for the first time.

I can't check this because I'm getting the `{"error":"Requires a valid JSON object value.","message":"Requires a valid JSON object value.","code":3}` error from the backend when I try to update.

> The login error is wiped due to a page reload. See attached GIF.
> Bad layout on Configuration page. See image below.
> Missing values on the configuration page (although they are available in the exported YAML file).
> Bad layout of warnings. See image (although may be fixed with what @novabyte said above). I think it is best to combine all warnings into one giant yellow box?

Fixed.

> When filtering storage by user ID, and then updating a given object, the view is punted back to the storage listing view and have forgotten the filtered user ID. Super annoying. I think this may be fixed with better usage of History API rather than href.location.

Yup, fixed.

mofirouz:
I can't test your final commit because I get this error when running `yarn install && yarn build` or `yarn serve`:

```
/Users/mo/Documents/go/src/github.com/heroiclabs/nakama/console/ui/src/routes/storage/index.tsx
TypeScript error: Type 'string | string[]' is not assignable to type 'string'.
  Type 'string[]' is not assignable to type 'string'.  TS2322

    73 |     if(query.user_id)
    74 |     {
  > 75 |       (document.getElementById('user_id') as HTMLInputElement).value = query.user_id;
       |       ^
    76 |     }
    77 |     this.props.fetchManyRequest(query);
    78 |   }
```

mofirouz:
Also what is the value you are trying to save when getting: 

```{"error":"Requires a valid JSON object value.","message":"Requires a valid JSON object value.","code":3}
```

Because yesterday I tried to update a pre-existing value and it worked. Are you trying to save an empty value field because it should always be saved as `{}`.

revnode:
@mofirouz Typescript doesn't always show the errors while running it, fixed. As far as what I'm pushing for `value`, I just change an integer from like a 1 to a 2, nothing major, and it spits out that error every time.

mofirouz:
@revnode I think we need to make a small tweak to the way CSV exports are done to make it follow RFC4180. Instead of escaping double quotes like this `\"`, RFC4180 says that double quotes should be escaped with another double quote (i.e. `""`). 

Let me know if this is difficult to do and I'll do some magic on the server to handle this, but I prefer we stick to RFC4180. Also, I don't think you need to put each field inside a `"`.

mofirouz:
To quote the godocs:

```
Fields which start and stop with the quote character " are called
quoted-fields. The beginning and ending quote are not part of the
field.

The source:

 normal string,"quoted-field"

results in the fields

 {`normal string`, `quoted-field`}

Within a quoted-field a quote character followed by a second quote
character is considered a single quote.

 "the ""word"" is true","a ""quoted-field"""

results in

 {`the "word" is true`, `a "quoted-field"`}

Newlines and commas may be included in a quoted-field

 "Multi-line
 field","comma is ,"

results in

 {`Multi-line
 field`, `comma is ,`}
```

PR #330, opened 2019-04-11
Console Status and Users Panels

mofirouz:


mofirouz:
A few observed issues:

- Disallow editing/deleting `0000-...` user. Actually don't allow clicking into the user detail view from the listing. Remove the "delete" button from the dashboard.
- Clicking on Wallet tab in user details view, produces the error attached in the picture.
- **Extra important** When clicking into the User Detail view, the dashboard automatically calls the "export" endpoint. This endpoint requires significant work from the server and should only be invoked on user's request to export data.
- The graphs in the Status view only show the current value plotted. They should accumulate a rolling 30min view of the graph. Everytime the server is polled for data, it should be added to the graph's plot data and graph updated.
- Bonus point: Do you think it's possible to not re-render the entire graph, just add the data point? At the moment, on re-render the entire animation kicks in and it is pretty jarring.
- If the server has crashed/not reachable, please redirect to the login view. Alternatively show a message somewhere that the server is unable to be reached.

revnode:
@mofirouz 
> Actually don't allow clicking into the user detail view from the listing. Remove the "delete" button from the dashboard.
I assume this is only for the 0000... user and not everybody else?
> Extra important When clicking into the User Detail view, the dashboard automatically calls the "export" endpoint. This endpoint requires significant work from the server and should only be invoked on user's request to export data.
All of the data in the export object is used in the panel. So it's either that one hit (with no hit for actual export) or multiple hits for each tab and then an additional hit for export. Your call.
> The graphs in the Status view only show the current value plotted. They should accumulate a rolling 30min view of the graph. Everytime the server is polled for data, it should be added to the graph's plot data and graph updated.
That's how it's done, it tracks history. If you're running this locally, there won't be much variation since it's local.
> Bonus point: Do you think it's possible to not re-render the entire graph, just add the data point? At the moment, on re-render the entire animation kicks in and it is pretty jarring.
I can turn off the animation? That should fix the jarring issue.

novabyte:
Hey @revnode 

> I assume this is only for the 0000... user and not everybody else?

Yes please. The system user ID is represented as the nil UUID who's string representation is the 0000... (zeroed value). Also if it's simple lets also prevent navigation to that particular user with the router as well. If navigation is made to that user we should just fallback on the users list view.

> All of the data in the export object is used in the panel. So it's either that one hit (with no hit for actual export) or multiple hits for each tab and then an additional hit for export. Your call.

I think the problem is that the export operation is "expensive" as a database task compared with the rest of the console API and is intended to be infrequent. I'll have a chat with @mofirouz and we'll get back to you.

> That's how it's done, it tracks history. If you're running this locally, there won't be much variation since it's local.

I think the rolling view of history should represent a one hour view so if we capture data points every 15 seconds we'd need to maintain up to 240 data points per graph. Does the history track all data points or does it accumulate and then rollover the 1 hour window?

> I can turn off the animation? That should fix the jarring issue.

I think if it's not easy to turn off the animation effects after first view load then lets just disable the animation fully.

Issue #242, opened 2018-09-26
Make it's possible to create group for more than 100 members and change default value via .yml config


Suvitruf:
Right now `100` is hardcoded here:

https://github.com/heroiclabs/nakama/blob/bb21e62f1043783ec6a0c3576ed05f0b997259d4/server/runtime_go_nakama.go#L1133

And, probably, here:

https://github.com/heroiclabs/nakama/blob/bb21e62f1043783ec6a0c3576ed05f0b997259d4/server/runtime_go_nakama.go#L1133

Please, make it's possible to:
1) Setup default value via .yml config.
2) Create group for more than 100 members.

zyro:
This is resolved in dd71a09b488b28f21270fd38e1a052ad042c7d78 and will be part of the next release. The server will allow any large value for the max count field.

PR #334, opened 2019-04-20
Set max group users count via config file

Suvitruf:
https://github.com/heroiclabs/nakama/issues/242

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -43,6 +43,7 @@ type Config interface {
 	GetSocket() *SocketConfig
 	GetDatabase() *DatabaseConfig
 	GetSocial() *SocialConfig
+	GetGroups() * GroupsConfig
 	GetRuntime() *RuntimeConfig
 	GetMatch() *MatchConfig
 	GetTracker() *TrackerConfig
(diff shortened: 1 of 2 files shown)


novabyte:
@Suvitruf I wonder if this might be better solved with a more suitable default value. What max group size do you use in your games?

Suvitruf:
Actually, we we want it to be unlimited, currently we are using 1000000. We have talked with you about it with in gitter. We don't request group members list, so for our case it's ok (:

GroovyAntoidDev:
I do wonder if there are performance considerations for bigger max group values.

If there is a large performance sacrifice, then it might be worth keeping a hard cap so people would not accidentally mess up their server unknowingly. (Or at least put a comment in the default config file)

Suvitruf:
> I do wonder if there are performance considerations for bigger max group values. (Or else it would not be hardcoded at 100 in the first place).
> 
> If there is a large performance sacrifice past certain thresholds, then it might be worth keeping a hard cap so people would not accidentally mess up their server unknowingly. (Or at least consider putting a comment in the default config file)

If you are requesting members list, there will be performance impact. But we don't do it. We only requesting total members count.

zyro:
After a lot of thought I think this should be handled differently to cover every use case:
* The concept of a max group user count should stay. It's a useful feature and the current implementation is the best way to handle it on top of the current schema, or if it's not needed it can be set to a very high value.
* Allow arbitrary max user count to be set at group creation time. This is already partially possible through the Go/Lua runtime (we'd need to remove the hard maximum) but it will probably be good to expose it to client group creation as well. Before hooks can ensure valid values are set here.
* Introduce pagination for group user listings, so performance of that operation is not directly affected by max group size.

The last point is the most important one. Once we have that covered there's no reason to cap the upper value allowed on max group sizes.

PR #350, opened 2019-06-11
Bug fixes for tournament end time schedules.

mofirouz:


server/core_tournament.go
```diff
@@ -222,7 +217,26 @@ WHERE duration > 0 AND start_time >= $1 AND end_time <= $2 AND category >= $3 AN
...
+	if endTime == 0 {
+		query += " AND end_time = $4"
+		params = append(params, time.Unix(0, 0).UTC())
+	} else if int64(endTime) < now.Unix() {
+		// if end time is set explicitly in the past
+		query += " AND end_time <= $4"
+		params = append(params, time.Unix(int64(endTime), 0).UTC())
+	} else {
+		// if end time is in the future, return both tournaments that end in the future as well as the ones that never end.
+		query += " AND (end_time <= $4 OR end_time = '1970-01-01 00:00:00 UTC')"
```
mofirouz: Open question:

Should I update the query to something like this:

```
query += " AND ((end_time >= now() AND end_time <= $4) OR end_time = '1970-01-01 00:00:00 UTC')"
```

This would mean any end_time between now, and in the future - excluding tournaments that ended previously. I don't think this is needed because we are doing that check in Nakama anyways.

data/modules/runonce_check.lua
```diff
@@ -13,17 +13,17 @@
...
+--  Test run_once function calls at server startup.
+----]]
+--
+--nk.run_once(function(context)
+--  assert(context.execution_mode, "run_once")
+--end)
+--
+--nk.run_once(function(context)
+--  error("Should not be executed.")
+--end)
```
zyro: Should these changes be committed?

data/modules/tournament.lua
```diff
@@ -1,98 +1,98 @@
...
+--  nk.tournament_delete(args.tournament_id)
+--end
+--nk.register_rpc(delete_tournament, "clientrpc.delete_tournament")
+--
+--local function addattempt_tournament(_context, payload)
+--  local args = nk.json_decode(payload)
+--
+--  nk.tournament_add_attempt(args.tournament_id, args.owner_id, args.count)
+--end
+--nk.register_rpc(addattempt_tournament, "clientrpc.addattempt_tournament")
```
zyro: Same as above, should these changes be committed?

server/core_tournament.go
zyro: I think it's fine as it is.

server/leaderboard_scheduler.go
```diff
@@ -229,12 +229,22 @@ func (ls *LocalLeaderboardScheduler) Update() {
 	}
 	if endActiveDuration > -1 {
 		ls.logger.Debug("Setting timer to run end active function", zap.Duration("end_active", endActiveDuration), zap.Strings("ids", ls.nearEndActiveIds))
+
+		if ls.endActiveTimer != nil {
+			ls.endActiveTimer.Stop()
```
zyro: I think this is already done above?

server/leaderboard_scheduler.go
```diff
@@ -229,12 +229,22 @@ func (ls *LocalLeaderboardScheduler) Update() {
...
+
 		ls.endActiveTimer = time.AfterFunc(endActiveDuration, func() {
 			ls.invokeEndActiveElapse(time.Unix(earliestEndActive-1, 0).UTC())
 		})
 	}
 	if expiryDuration > -1 {
 		ls.logger.Debug("Setting timer to run expiry function", zap.Duration("expiry", expiryDuration), zap.Strings("ids", ls.nearExpiryIds))
+
+		if ls.expiryTimer != nil {
+			ls.expiryTimer.Stop()
```
zyro: Same as above, I think this is already done earlier in the function?

PR #355, opened 2019-06-27
Add scripts to generate DigitalOcean Marketplace image

joenash:
This PR adds the scripts to generate the [Nakama app](https://marketplace.digitalocean.com/apps/nakama) on the DigitalOcean marketplace. 

This script uses binary installs for Nakama and CockroachDB, with the URL for the binary set as a variable in the script. This could be automated, using a webhook on the repository for the [release event](https://developer.github.com/v3/activity/events/types/#releaseevent), to grab the latest Nakama release and build the DO image.


joenash:
@novabyte Wasn't sure if this was the most appropriate place/way to add the DO files to the repo, let me know what you think and if they need moving

novabyte:
Thanks @joenash. The build folder is a good place to keep these utilities which we can use for releases. I'll take a look and leave a review.

Issue #323, opened 2019-04-10
Add an option to ignore full groups when listing groups


casperkl:
## Description
We want to show the users a list of possible groups they can join, for this we want to prevent returning a list of full groups. An option to ignore groups that have reached their member count would help with this.
For searches we would want to include full groups, so an option would be ideal.

PR #364, opened 2019-07-25
Add an option to ignore full groups when listing groups

emmrk:
This is a take on issue https://github.com/heroiclabs/nakama/issues/323

Adds an option to exclude full groups from appearing on the list of all groups. Off by default to preserve existing behavior.

To enable, add an `omit_full=true` option to the API request:
```
curl "http://127.0.0.1:7350/v2/group?limit=100&omit_full=true" -H 'Authorization: Bearer <TOKEN>'
```


api/api.proto
```diff
@@ -498,6 +498,8 @@ message ListGroupsRequest {
   string cursor = 2;
   // Max number of groups to return. Between 1 and 100.
   google.protobuf.Int32Value limit = 3;
+  // Whether to include full groups in the listing.
+  google.protobuf.BoolValue omit_full = 4;
```
mofirouz: I think we can just use `bool` instead of `google.protobuf.BoolValue` since the absence of that value would be `false` by default and thus would be the old/default behaviour of the server.

server/core_group.go
```diff
@@ -996,12 +996,16 @@ func ListGroups(ctx context.Context, logger *zap.Logger, db *sql.DB, name string
...
+
 	if name == "" {
 		query = `
 SELECT id, creator_id, name, description, avatar_url, state, edge_count, lang_tag, max_count, metadata, create_time, update_time
 FROM groups
-WHERE disable_time = '1970-01-01 00:00:00 UTC'
+WHERE
+	(disable_time = '1970-01-01 00:00:00 UTC')
+AND
+	((NOT $2) OR (edge_count < max_count))
```
mofirouz: I think this and subsequent `OR (edge_count < max_count)` clauses significantly change performance profile of the query. Have you done an analysis on the query? What does `ANALYSE ....` look like?

I ask, because the originally query was looking at indexed fields, and also was a key-value look up (as much as it could), and also the indexes were created in such a fashion to allow the database to fast-skip as many records as possible. However, the new query adds a range query that doesn't allow the database to skip records.

api/api.proto
mofirouz: I think this can be a `bool` rather than `google.protobuf.BoolValue` since the absence of the value would be `false` and therefore the old/default behaviour of the server.

server/core_group.go
mofirouz: I think this addition (`OR (edge_count < max_count)`) to the query will have a significant impact on the performance profile of the query. 

I ask, because the original query was a partial k-v look up and was hitting indexed fields as much as possible. Additionally the use of `OR` and range queries does not allow the database to fast-skip records on disk.

Have you performed `ANALYSIS ... ` on this query?

api/api.proto
emmrk: Right, there's no need to use the protobuf wrapper, changed to vanilla `bool`

server/core_group.go
emmrk: Yes, I've seen the query analysis, and of course its performance may only be worse than it used to be without this additional filter. Although it doesn't seem more than marginally so when I try running these queries on a database with 10k entries in the `groups` table. Would that be in any way indicative of the real use conditions?

Anyway, it'll only matter for the queries with `omit_full=true` and after all the other filters on indexed fields have been applied. Unfortunately, I do not know much about the typical loads, so it's hard to judge whether an additional index is worth it for this kind of a query.

I've also studied the CockroachDB documentation, and it doesn't seem to say anything about issues with skipping ranges when using this or that SQL feature. Could you please explain what you meant in the second paragraph?

zyro:
The approach to this feature will need some further thought, the query performance of this implementation as it is needs improvement before it can be merged for mainstream use. We'll revisit this in the future.

PR #367, opened 2019-07-31
Switch prometheus metrics to use labels instead of a prefix

joshgarnett:
Prometheus has the concept of labels, which allow you to add sub fields to a metric name.  This makes it easy to generate queries for a metric across a number of labels.  In the case of nakama, this will allow for the building of single dashboard, which can render metrics across any number of nodes.

With this change, the metric nakama_b26f_local_nakama_socket_ws_server_elapsed_time_bucket{le="0.0"} would become nakama_socket_ws_server_elapsed_time_bucket{le="0.0", node_name="nakama_b26f", namespace="local"}

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -29,7 +29,6 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
 	"go.uber.org/zap"
-	"strings"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


joshgarnett:
If users are relying on the old behavior, we might want to create a setting that lets them toggle between the behaviors.  That said, this change makes the metrics much more idiomatic for prometheus and likely be something existing users will want to move to.

zyro:
@joshgarnett Awesome! I'll add a note to the changelog to highlight this change and we'll call it out in the release notes.

PR #390, opened 2019-10-28
fix time conversion for time based metrics

kevhong:
**Background**
- Conversion rate from nanoseconds to milliseconds is 1e6 not 1k
- https://www.convertunits.com/from/nanosecond/to/millisecond

**Change**
- Updated in  api (MetricsAPITimeSpentMsec) and websocket (MetricsSocketWsTimeSpentMsec)
- MetricsRtapiTimeSpentMsec does not seem be to used anymore, so there are no updates for it


zyro:
@kevhong Thanks!

PR #391, opened 2019-10-28
Fix for 8 char pwd error message.

XavierOlland:
Message was "Password must be longer than 8 characters." when actually 8 character is the minimum.


zyro:
@XavierOlland Thanks!

PR #392, opened 2019-10-28
Fix in Documentation

thetemplar:
Minor fix in the documentation (which cost me way to long to figure it out)


thetemplar:
While we are on it: README.md also needs a minor adjustment.
`-i` is not suggested anymore to use, see [golang/go/issues/24674#issuecomment-379143645](https://github.com/golang/go/issues/24674#issuecomment-379143645) and `--trimpath` needs to be added :)

PR #396, opened 2019-11-09
Add a note to manually fix nakama-commons version

harelba:


harelba:
no need for this, added by mistake.

PR #399, opened 2019-11-19
Fix ListNotifications not using the user supplied cursor

jdetter:
This was a fix made by Zeke (bfops). This fixes the bug where ListNotifications will always start at the beginning of your notifications list because the input cursor isn't assigned properly.


zyro:
@jdetter Thanks!

PR #409, opened 2020-01-09
Facebook Instant Game authentication

emmrk:
Add support for authenticating with the Nakama server by providing Facebook Instant Game signed player information.

Depends on https://github.com/heroiclabs/nakama-common/pull/1

https://developers.facebook.com/docs/games/instant-games/sdk/fbinstant6.3#signedplayerinfo


apigrpc/apigrpc.swagger.json
```diff
@@ -279,6 +279,53 @@
...
+        "parameters": [
+          {
+            "name": "body",
+            "description": "The Facebook Instant account details.",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/apiAccountFacebookInstantGame"
+            },
+          {
```
Dimon4eg: missing closing bracket?

migrate/sql/20180103142001_initial_schema.sql
```diff
@@ -18,28 +18,29 @@
...
-    disable_time  TIMESTAMPTZ   DEFAULT '1970-01-01 00:00:00 UTC' NOT NULL
+    lang_tag                  VARCHAR(18)   DEFAULT 'en',
+    location                  VARCHAR(255), -- e.g. "San Francisco, CA"
+    timezone                  VARCHAR(255), -- e.g. "Pacific Time (US & Canada)"
+    metadata                  JSONB         DEFAULT '{}' NOT NULL,
+    wallet                    JSONB         DEFAULT '{}' NOT NULL,
+    email                     VARCHAR(255)  UNIQUE,
+    password                  BYTEA         CHECK (length(password) < 32000),
+    facebook_id               VARCHAR(128)  UNIQUE,
+    facebook_instant_game_id  VARCHAR(128)  UNIQUE,
```
zyro: We should add this change as a new migration file. The migration tool checks the schema based on which migration files it knows have been applied, rather than comparing tables/indexes.

If we don't do this as a new migration users with old databases that upgrade to a newer Nakama version will get errors when trying to use FBIG auth.

zyro: I'm also assuming the format/length of FB and FBIG IDs is the same. If not we'll need to adjust the column type.

server/core_authenticate.go
```diff
@@ -384,6 +384,76 @@ func AuthenticateFacebook(ctx context.Context, logger *zap.Logger, db *sql.DB, c
 	return userID, username, true, nil
 }
 
+func extractInstantGameID(signedPlayerInfo string, appSecret string) (facebookInstantGameID string, err error) {
```
zyro: If this is the function that will handle validating the token and potentially requesting the user's profile (and friends) from Facebook services then it should live in the `social` package.

server/api_authenticate.go
```diff
@@ -341,6 +341,65 @@ func (s *ApiServer) AuthenticateFacebook(ctx context.Context, in *api.Authentica
...
+		username = generateUsername()
+	} else if invalidCharsRegex.MatchString(username) {
+		return nil, status.Error(codes.InvalidArgument, "Username invalid, no spaces or control characters allowed.")
+	} else if len(username) > 128 {
+		return nil, status.Error(codes.InvalidArgument, "Username invalid, must be 1-128 bytes.")
+	}
+
+	create := in.Create == nil || in.Create.Value
+
+	dbUserID, dbUsername, created, err := AuthenticateFacebookInstantGame(ctx, s.logger, s.db, s.socialClient, s.config.GetSocial().FacebookInstantGame.AppSecret, in.Account.SignedPlayerInfo, username, create)
```
zyro: Is it possible to do a friends import operation similar to standard Facebook login? If that's the case we should support it, see https://github.com/heroiclabs/nakama/blob/master/server/core_authenticate.go#L627 for how it's done when handling Facebook auth.

apigrpc/apigrpc.proto
```diff
@@ -162,6 +162,22 @@ service Nakama {
     };
   }
 
+  // Authenticate a user with a Facebook Instant Game token against the server.
+  rpc AuthenticateFacebookInstantGame (api.AuthenticateFacebookInstantGameRequest) returns (api.Session) {
```
zyro: Also add the corresponding link/unlink operations.

server/runtime.go
```diff
@@ -716,6 +723,10 @@ func NewRuntime(logger, startupLogger *zap.Logger, db *sql.DB, jsonpbMarshaler *
 		allBeforeReqFunctions.beforeAuthenticateFacebookFunction = goBeforeReqFunctions.beforeAuthenticateFacebookFunction
 		startupLogger.Info("Registered Go runtime Before function invocation", zap.String("id", "authenticatefacebook"))
 	}
+	if goBeforeReqFunctions.beforeAuthenticateFacebookInstantGameFunction != nil {
```
zyro: Also add support for Lua before hook, see: https://github.com/heroiclabs/nakama/blob/master/server/runtime_lua.go#L229

server/runtime.go
```diff
@@ -1131,6 +1145,10 @@ func NewRuntime(logger, startupLogger *zap.Logger, db *sql.DB, jsonpbMarshaler *
 		allAfterReqFunctions.afterAuthenticateFacebookFunction = goAfterReqFunctions.afterAuthenticateFacebookFunction
 		startupLogger.Info("Registered Go runtime After function invocation", zap.String("id", "authenticatefacebook"))
 	}
+	if goAfterReqFunctions.afterAuthenticateFacebookInstantGameFunction != nil {
```
zyro: Also add support for Lua after hook, see: https://github.com/heroiclabs/nakama/blob/master/server/runtime_lua.go#L691

server/api_authenticate.go
emmrk: It is not, but the Instant App may query FB for the list of "connected players" and establish "friendships" by issuing independent requests to nakama:

> `getConnectedPlayersAsync( )`
> 
> Fetches an array of ConnectedPlayer objects containing information about active players (people who played the game in the last 90 days) that are connected to the current player.

apigrpc/apigrpc.swagger.json
emmrk: thanks! fixed

server/core_authenticate.go
emmrk: It doesn't request anything from FB, only validates the token and extracts the Facebook Instant Game player ID from it. 

Moved to `social` nonetheless.

apigrpc/apigrpc.proto
emmrk: Done

server/runtime.go
emmrk: Done

server/runtime.go
emmrk: Done

apigrpc/apigrpc.swagger.json
```diff
@@ -540,6 +588,34 @@
...
+              "$ref": "#/definitions/apiAccountFacebookInstantGame"
+            }
+          }
+        ],
+        "tags": [
+          "Nakama"
+        ]
+      }
+    },
     "/v2/account/link/gamecenter": {
```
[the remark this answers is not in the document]
zyro: All link/unlink endpoints require authentication. If this isn't reflected in the swagger file then that's a bug - or perhaps a limitation of the generator? @novabyte can you remember?

server/api_unlink.go
```diff
@@ -312,6 +315,75 @@ AND ((custom_id IS NOT NULL
...
+	if err != nil {
+		s.logger.Info("Could not authenticate Facebook Instant Game profile.", zap.Error(err))
+		return nil, status.Error(codes.Unauthenticated, "Could not authenticate Facebook Instant Game profile.")
+	}
+
+	query := `UPDATE users SET facebook_instant_game_id = NULL, update_time = now()
+WHERE id = $1
+AND facebook_instant_game_id = $2
+AND ((custom_id IS NOT NULL
+      OR google_id IS NOT NULL
```
zyro: Add `OR facebook_id IS NOT NULL` just above this line I think.

PR #415, opened 2020-03-30
Make token bearer auth the default in console.proto

sesposito:


novabyte:
Nice one @sesposito 👍

Issue #402, opened 2019-12-12
Request: Add Demote Operation In Group


ernesernesto:
## Description
As explained in https://forum.heroiclabs.com/t/handling-demote-for-the-receiving-user/383/2 we want to have a way from the client to easily demote user in a group. While current method to demote is available via removing from group and then re-added it, from the pov of the user being demoted will received two messages, kicked and then added, which will cause confusion to that user. It would be better to handle it if we have message code for Demotion and a way for a client to demote user

## Context
- [x] Unity

## Your Environment
- Nakama: 2.8.0

sesposito:
Resolved in #445

PR #435, opened 2020-06-18
Add group member demote operation

emmrk:
Depends on definitions from https://github.com/heroiclabs/nakama-common/pull/3
Fixes https://github.com/heroiclabs/nakama/issues/402


sesposito:
Resolves by #445

PR #441, opened 2020-07-09
Use basic auth in metadata check for console authentication

### LINKED ISSUE #440: Proposal to use basic auth in headers instead of body for console API
The C# client code generator produces incorrect code from the console API because it doesn't know how to generate a method for the `/v2/console/authenticate` operation.

This happens because the operation in the `console.swagger.json` omits `BasicAuth` as a security option and opts to authenticate via a body param in lieu of the headers.

https://github.com/heroiclabs/nakama/blob/master/console/console.swagger.json#L737

The following would make the client generate a successful method:

(1) Remove the `consoleAuthenticateRequest` parameter.
(2) Use `BasicAuth` as the security option. 

Of course, the client code generator could be modified to accommodate this, but frankly it would make the code a lot more brittle. If this is something that is not a big deal to do on the backend, that would be better. But always up for thinking about alternatives.

lugehorsam:
Streamlines the authentication process for the Console API. By doing an initial authentication via headers/metadata, the Console SDK code generators can properly emit the `Authenticate` method by reading the `BasicAuth` security option in Console Swagger doc.

Closes #440


lugehorsam:
There is a change I needed to make on the frontend. Will reopen shortly.

lugehorsam:
Ready for review.

lugehorsam:
Putting a reminder here to myself to think about how this change to the console API will affect any current GUIs that our clients have built on top of the Console API.

lugehorsam:
Closing this PR because @zyro and I talked through a way to modify the code generator that makes a lot of sense.

PR #445, opened 2020-08-03
Implement go runtime structured logger

sesposito:
Add group member demote operation (from #435)


apigrpc/apigrpc.swagger.json
```diff
@@ -133,21 +163,6 @@
             "schema": {
               "$ref": "#/definitions/apiAccountCustom"
             }
-          },
```
zyro, line 136: I think you need to revert all these deletions in the swagger file manually.

PR #450, opened 2020-08-10
Add missing user_ids param from swagger group demote

sesposito:


apigrpc/apigrpc.swagger.json
```diff
@@ -1748,6 +1748,16 @@
             "in": "path",
             "required": true,
             "type": "string"
+          },
+          {
+            "name": "user_ids",
+            "description": "The users to demote.",
+            "in": "query",
+            "required": false,
```
lugehorsam, line 1756: I think this would be required, wouldn't it?

lugehorsam:
👍

PR #456, opened 2020-08-27
use logger.Info instead of logger.Printf in sample go runtime code

lugehorsam:
The sample code was not compiling due to `Printf` not existing on the logger. I just modified it to use `Info` instead.


lugehorsam:
I also added an rpc call for creating matches so users can see fully how custom match handlers are used.

zyro (approved):

zyro:
Good stuff! 👍

PR #457, opened 2020-08-27
Fix bugs in a couple of Lua nk module functions.

sesposito:
1) jwt_generate now correctly supports the RS256 algorithm.
2) rsa_sha256_hash now correctly decoding the private key argument.

### DIFF
--- a/server/runtime_lua_nakama.go
+++ b/server/runtime_lua_nakama.go
@@ -808,7 +808,8 @@ func (n *RuntimeLuaNakamaModule) jwtGenerate(l *lua.LState) int {
 	case "RS256":
 		signingMethod = jwt.SigningMethodRS256
 	default:
-		l.ArgError(3, "unsupported algo type - only allowed 'HS256', 'RS256'.")
+		l.ArgError(1, "unsupported algo type - only allowed 'HS256', 'RS256'.")
+		return 0
 	}
 
 	signingKey := l.CheckString(2)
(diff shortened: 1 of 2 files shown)


server/runtime_lua_nakama.go
```diff
@@ -1129,7 +1136,8 @@ func (n *RuntimeLuaNakamaModule) rsaSHA256Hash(l *lua.LState) int {
 		return 0
 	}
 
-	rsaPrivateKey, err := x509.ParsePKCS1PrivateKey([]byte(key))
+	block, _ := pem.Decode([]byte(key))
```
zyro, line 1139: We should observe the error here and return early if non-nil.

zyro (changes requested):

zyro:
Also in the algo selection here https://github.com/heroiclabs/nakama/pull/457/files#diff-ac71981a54f4a4b70786ee656dde0bcfR811 the default case should be an arg error for the 1st parameter (not the 3rd) and we should also return early there.

zyro:
@sesposito Couple of small changes otherwise looks nice! 👍

server/runtime_lua_nakama.go
sesposito: The second returned value of `pem.Decode` is not an error, but the remaining input if it has multiple pem formatted blocks. I'll check if the returned `block` is nil instead and raise an error if it is.

zyro: Excellent point, the check if it's `nil` is a good move then.

zyro (approved):

PR #460, opened 2020-09-10
Fix broken links to cloud offer in README

hatkyinc2:
Not sure if you want to change the text, but the links lead to 404, I assume this is the desired replacement link.
(Could also update the site in it's place..)


novabyte:
Nice catch. Thank you @hatkyinc2 🙏

PR #466, opened 2020-09-29
Add refresh tokens to authentication API - PREVIEW

michal-harish:
- refresh tokens have separate expiry and encryption key config - DONE
- refresh tokens are returned in Session object alongside the regular token - DONE
- refresh tokens can only be used with AuthenticateRefresh request - DONE **(provided the refresh_encryption_key is different from encryption_key)**
- `/v2/authenticate/refresh` plus Authorization header carrying the Bearer <refresh_token> returns a renewed token pair - DONE
- **Currently only added to AuthenticateDevice but I assume it will need adding to all other auth methods ?**
- **what about before and after auth hooks ?**
- **Also not implemented is the one-timeness of the refresh token - i think it will require some kind of state that keeps the track of already used tokens**
- **default expiry of the refresh tokens is currently 3600 sec but should be longer probably**


server/api.go
```diff
@@ -276,6 +276,31 @@ func (s *ApiServer) Healthcheck(ctx context.Context, in *empty.Empty) (*empty.Em
...
+		}
+
+		if len(auth) != 1 {
+			// Value of "authorization" or "grpc-authorization" was empty or repeated.
+			return nil, status.Error(codes.Unauthenticated, "Auth token invalid")
+		}
+
+		//TODO check re-use and FailPrecondition
+
+		userID, username, vars, exp, ok := parseBearerAuth([]byte(config.GetSession().RefreshEncryptionKey), auth[0])
```
zyro, line 297: I think instead of passing this token as part of the `Authorization` header we should make it part of the request body, and require the server key in the `Authorization` header like the other authentication endpoints.

server/api_authenticate.go
```diff
@@ -660,19 +661,44 @@ func (s *ApiServer) AuthenticateSteam(ctx context.Context, in *api.AuthenticateS
 	return session, nil
 }
 
+func (s *ApiServer) AuthenticateRefresh(ctx context.Context, in *api.AuthenticateRefreshRequest) (*api.Session, error) {
+	uid := ""
+	username := ""
+	var vars map[string]string
```
zyro, line 667: Rather than read `vars` from the token allow it to be passed through in the `AuthenticateRefresh` request body, gives us extra flexibility to replace them if needed.

zyro (changes requested):

zyro:
Merged under https://github.com/heroiclabs/nakama/pull/527.

Issue #461, opened 2020-09-19
Database credentials visible in logs


peterkingsbury:
## Description
Database credentials (username+password) visible in logs on startup at numerous points during startup.

This is less of a concern when Nakama is running isolated from other services, but when logs are shipped to external sources (e.g. Elastic or others) anyone with access to those external tools has access to database credentials.

### Assumptions
A database (I used PostgreSQL, but CockroachDb would be fine) with the setup indicated in the `.env` file.

### Steps
1. Start with a docker-compose file similar to [Install Docker Quickstart](https://heroiclabs.com/docs/install-docker-quickstart/).
2. Modify the docker-compose.yml file to obtain credentials from the environment:
```
...
    env_file:
      - ./.env
    entrypoint:
      - "/bin/sh"
      - "-ecx"
      - >
        /nakama/nakama migrate up --database.address ${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB} &&
        exec /nakama/nakama --name nakama1 --database.address ${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB} --console.password=${CONSOLE_PASSWORD}
...
```
3. Add an `.env` file with the following:
```
POSTGRES_USER=MYDBUSERNAME
POSTGRES_PASSWORD=MYSUPERSECRETPASSWORD
POSTGRES_HOST=MYDBHOST
POSTGRES_PORT=MYDBPORT
POSTGRES_DB=MYDB
CONSOLE_PASSWORD=MYCONSOLEPASSWORD
```
4. Start Nakama with `docker-compose up -d`.
5. Observe log output using `docker logs -f nakama`.
6. Compare log output to XHR `GET /v2/console/config?` in web console.

## Expected Result
Database credentials are obscured before being delivered to stdout/stderr (e.g. `MYDBUSERNAME:********@MYDBHOST:MYDBPORT/MYDB` instead of `MYDBUSERNAME:MYSUPERSECRETPASSWORD@MYDBHOST:MYDBPORT/MYDB`).

Comparing this with Config section in the web console, the console scrubs credentials (database password, console password) before sending to a UI client.

## Actual Result
Credentials clearly indicated as both part of the command-execution being sent to the container (unsure whether this can be fixed) as well as the Nakama log output itself:
```
[...]

PR #467, opened 2020-10-03
Fix for database credentials visible in logs

peterkingsbury:
Issue logged in #461 indicating passwords being output to logs during startup.

This PR uses [URL.Redacted()](https://golang.org/pkg/net/url/#URL.Redacted) to display the DSN when logged.

During migrations, DSNs are output using the alternate format.

During startup, a copy is made, then the copy is redacted and used for log output.


novabyte:
Thanks @peterkingsbury we'll take a look at this next week 🙏

zyro (approved):

PR #472, opened 2020-10-16
Return rank in lua tournament_records_haystack fn

### LINKED ISSUE #469: tournament_records_haystack funtion in Lua runtime does not return "rank" value
https://github.com/heroiclabs/nakama/blob/94611911afda28a7dec480a6adf9c4a3b2d3a801/server/runtime_lua_nakama.go#L5969

It seem the issues is in the function above.

sesposito:
Fixes #469

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project are documented below.
 The format is based on [keep a changelog](http://keepachangelog.com) and this project uses [semantic versioning](http://semver.org).
 
 ## [Unreleased]
-
+### Fixed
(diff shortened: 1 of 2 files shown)


novabyte:
@sesposito Nice catch. Can you add a note to the CHANGELOG as well. 👍

zyro:
@sesposito This line should be adjusted too https://github.com/heroiclabs/nakama/pull/472/files#diff-df30f09592ff0f9c092c2b7b3692365c85e4ceb502e1ee86b0e706e7891dbcbaL6002, from `10` to `11` to account for the extra field.

novabyte (approved):

PR #475, opened 2020-10-20
Add TournamentRecordsList and FriendList to core

### LINKED ISSUE #326: Listing friends in a Lua module
It would be nice if there was a built in way to be able to check a user's friends list within a lua module. There is a pretty easy work around for this but it requires writing a SQL query. 

## Use case

In our game we are implementing a party system where you can queue into the matchmaker with a party and we want to restrict it so that you can only send nakama notifications/push notifications to players who are on your friends list.

## Work around

For now you can use a custom SQL query to get all of the friends for a given player, but it would be much better if there was built-in functionality for listing friends, adding friends, removing friends, ect.

### LINKED ISSUE #470: [Feature] Add runtime equivalent to LeaderboardRecordsList for tournaments
Seems to be only missing on the runtime as the client has this api

https://forum.heroiclabs.com/t/leaderboardrecordslist-for-tournaments/1045

sesposito:
`ListTournamentRecords` logic moved from api to core so it can be reused.
Expose `ListTournamentRecords` to Go and Lua runtimes.
Add FriendsList function to Go and Lua runtimes.

Fixes #326 #470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,12 @@ The format is based on [keep a changelog](http://keepachangelog.com) and this pr
 - Event contexts now contain user information for external events.
 - Expose more metrics for socket activity.
 - New [Docker release](https://hub.docker.com/repository/docker/heroiclabs/nakama-dsym) of the server with debug symbols enabled.
+- Added `TournamentRecordsList` and `ListFriends` functions to the Go Runtime.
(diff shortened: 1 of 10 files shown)


zyro (approved):

zyro:
@sesposito Before you merge make sure you update the vendored deps with the new nakama-common version that adds tournament record listing to the Go runtime Nakama module. :+1:

zyro (approved):

PR #477, opened 2020-10-21
Update protoc-gen-angular go.mod

sesposito:
Allow the angular generator protoc plugin to be installed using go get.


mofirouz (approved):
love it ❤️

PR #481, opened 2020-10-28
Nakama3 console auth

michal-harish:


server/console_authenticate.go
```diff
@@ -16,26 +16,119 @@ package server
...
 
 	"github.com/heroiclabs/nakama/v2/console"
 )
 
+var validRoles = map[string]bool{
+
+}
+
+type ConsoleTokenClaims struct {
+	Username  string           `json:"usn,omitempty"`
```
mofirouz, line 36: We should add `email` here too.

server/console_authenticate.go
```diff
@@ -16,26 +16,119 @@ package server
...
+	key := []byte(s.config.GetConsole().SigningKey)
+	signedToken, _ := token.SignedString(key)
+	uname, role, exp, _ := parseConsoleToken(key, signedToken)
+	fmt.Println(uname, role, exp)
+
+	return &console.ConsoleSession{Token: signedToken}, nil
+}
+
+func (s *ConsoleServer) lookupConsoleUser(ctx context.Context, unameOrEmail, password string) (console.UserRole, error) {
+	query := "SELECT role, password, disable_time FROM console_users WHERE username = $1 OR eamil = $1"
```
mofirouz: I think select `username, email` as well on the return query

mofirouz (changes requested):
Couple of minor adjustments, otherwise looks good to me.

mofirouz:
This LGTM - Please add the security interceptor that checks the `role` and then I think we can merge this.

server/console.go
```diff
@@ -112,7 +116,17 @@ func StartConsoleServer(logger *zap.Logger, startupLogger *zap.Logger, db *sql.D
...
 		}),
 	)
-	if err := console.RegisterConsoleHandlerServer(ctx, grpcGateway, s); err != nil {
+
+	dialAddr := fmt.Sprintf("127.0.0.1:%d", config.GetConsole().Port-3)
+	if config.GetConsole().Address != "" {
+		dialAddr = fmt.Sprintf("%v:%d", config.GetConsole().Address, config.GetConsole().Port-3)
+	}
+	dialOpts := []grpc.DialOption{
+		//TODO (mo, zyro): Do we need to pass the statsHandler here as well?
```
mofirouz, line 125: no stats handler needed here.

server/console.go
```diff
@@ -235,15 +222,52 @@ func consoleInterceptorFunc(logger *zap.Logger, config Config) func(context.Cont
...
 		}
 
-		if !checkAuth(config, auth[0]) {
+		if ctx, ok = checkAuth(ctx, config, auth[0]); !ok {
 			return nil, status.Error(codes.Unauthenticated, "Console authentication invalid.")
 		}
-
+		role := ctx.Value(ctxConsoleRoleKey{}).(console.UserRole)
+		forbidden := false
+		switch role {
```
mofirouz, line 230: would be good to have 4 `const maps[string]interface{}{{"/nakama.console.Console/AddUser": nil},...}` types, and avoid the nested switch statement

mofirouz (commented):
small comments, otherwise LGTM

Issue #480, opened 2020-10-25
Incorrect rank values in "list leaderboard records around owner"


AGulev:
## Description
Some records don't have `rank` field in "list leaderboard records around owner" request. Records that have this fields are wrong (they are fine after restar, but if I create a couple of new users and add scores, it breaks)

## Steps to Reproduce
1. Create leaderboard
```
local nk = require("nakama")

local authoritative = false
local sort = "desc"
local operator = "best"
local reset = "0 0 1 * *" -- monthly
local metadata = {}

nk.leaderboard_create("leisure_level", authoritative, sort, operator, reset, metadata)
```
2. Create 20 users and submit random scores
3. Login as user in the middle of leaderboard
4. Request top 10 records using regular request List Leaderboard Records
`HTTP	GET	http://127.0.0.1:7350/v2/leaderboard/leisure_level?limit=10`
5. Request 7 records using request List leaderboard records around owner:
`HTTP	GET	http://127.0.0.1:7350/v2/leaderboard/leisure_level/owner/2064feff-eab6-4327-a896-1b5f05ed6c83?limit=7`
6. Compare responses

```
DEBUG:SCRIPT: HTTP	GET	http://127.0.0.1:7350/v2/leaderboard/leisure_level?limit=10
DEBUG:SCRIPT: DATA	nil
DEBUG:SCRIPT: {
   "records":[
      {
         "leaderboard_id":"leisure_level",
         "owner_id":"7797a3cd-bfa0-4a5e-8248-33ffbc7247f7",
         "username":"PxcfsZicfI",
         "score":"909",
         "num_score":1,
         "metadata":"{}",
         "create_time":"2020-10-25T16:04:01Z",
         "update_time":"2020-10-25T16:04:01Z",
         "expiry_time":"2020-11-01T00:00:00Z",
         "rank":"1",
         "max_num_score":1000000
      },
      {
         "leaderboard_id":"leisure_level",
         "owner_id":"e52c55a9-71df-40e4-994a-d4b2add81fd9",
         "username":"rGQHTCMJIy",
         "score":"906",
         "num_score":1,
         "metadata":"{}",
         "create_time":"2020-10-25T17:49:36Z",
         "update_time":"2020-10-25T17:49:36Z",
         "expiry_time":"2020-11-01T00:00:00Z",
         "rank":"2",
         "max_num_score":1000000
      },
      {
         "leaderboard_id":"leisure_level",
[...]

PR #482, opened 2020-10-30
Fix leaderboard rank sorting

4726:
I believe this also fixes https://github.com/heroiclabs/nakama/issues/480


zyro:
Nice find, thanks @4726!

PR #486, opened 2020-11-05
Nakama3 console auth rebase (#485)

michal-harish:
* console auth - authentication, validation and premission checking based on roles
* console grpc gateway - revert from direct to loopback forward ing from gateway to grpc server


sesposito:
https://github.com/heroiclabs/nakama/pull/486/files#diff-12b6715f24b4d9f749659362ac764d5b7b2720a04d69b4d5bcb86eee9d611c74R109 

@michal-harish can you please add the `EmitUnpopulated: true` option here?

Otherwise LGTM 👍

PR #491, opened 2020-11-12
Fix tournament related bugs

sesposito:
- Fix an issue that would cause an error when trying to update a tournament record with invalid data.
- Fix some issues around listing tournaments.
- Fix an issue that would prevent the insertion of a record in a tournament with no scheduled reset and end time.

### DIFF
--- a/server/core_tournament.go
+++ b/server/core_tournament.go
@@ -396,11 +396,11 @@ func TournamentRecordWrite(ctx context.Context, logger *zap.Logger, db *sql.DB, 
 	default:
 		if leaderboard.SortOrder == LeaderboardSortOrderAscending {
 			// Lower score is better.
-			opSQL = "score = div((leaderboard_record.score + $5 - abs(leaderboard_record.score - $5)), 2), subscore = div((leaderboard_record.subscore + $6 - abs(leaderboard_record.subscore - $6)), 2)"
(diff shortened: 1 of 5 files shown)


server/core_tournament.go
```diff
@@ -453,49 +453,58 @@ func TournamentRecordWrite(ctx context.Context, logger *zap.Logger, db *sql.DB,
...
 
 			return nil
 		}); err != nil {
 			if err == ErrTournamentWriteMaxNumScoreReached || err == ErrTournamentMaxSizeReached {
 				logger.Info("Aborted writing tournament record", zap.String("reason", err.Error()), zap.String("tournament_id", tournamentId), zap.String("owner_id", ownerId.String()))
-			} else {
-				logger.Error("Could not write tournament record", zap.Error(err), zap.String("tournament_id", tournamentId), zap.String("owner_id", ownerId.String()))
 			}
+
+			logger.Error("Could not write tournament record", zap.Error(err), zap.String("tournament_id", tournamentId), zap.String("owner_id", ownerId.String()))
```
zyro, line 507: I think the `else {` clause here should be kept. If we don't keep it then an error of `ErrTournamentMaxSizeReached` will print both the info and the error message and return early.

zyro (changes requested):

zyro (approved):

PR #500, opened 2020-11-24
Add console match listing API

sesposito:
Add tick_rate and handler_name to returned fields in the listing and make them part of the bleve index.


server/runtime.go
```diff
@@ -243,7 +243,7 @@ func (e RuntimeExecutionMode) String() string {
 }
 
 type RuntimeMatchCore interface {
-	MatchInit(presenceList *MatchPresenceList, deferMessageFn RuntimeMatchDeferMessageFunction, params map[string]interface{}) (interface{}, int, error)
+	MatchInit(module string, presenceList *MatchPresenceList, deferMessageFn RuntimeMatchDeferMessageFunction, params map[string]interface{}) (interface{}, int, error)
```
zyro, line 246: The module name should be known at match core creation time, before `MatchInit`. I suggest we pass it through the `NewRuntimeGoMatchCore` and `NewRuntimeLuaMatchCore` functions.

Additionally the match core might have a `GetHandlerName() string` function to return it, saves having to thread it through the match registry's `NewMatch` I think?

zyro (changes requested):

server/match_handler.go
```diff
@@ -106,11 +106,12 @@ type MatchHandler struct {
 
 	// Match state.
 	state interface{}
+
+	TickRate int
```
zyro, line 110: Just noticed we already have this field above named `Rate`. Can we simplify to just reuse that one?

server/match_handler.go
```diff
@@ -142,9 +143,10 @@ func NewMatchHandler(logger *zap.Logger, config Config, sessionRegistry SessionR
 		PresenceList:   presenceList,
 		core:           core,
```
zyro, line 144: Let's also flip the `core` field to be exported as `Core` here, it will be needed later.

zyro (changes requested):

zyro (approved):

PR #504, opened 2020-11-26
Add advanced filters to console storage listing.

michal-harish:


console/console.proto
```diff
@@ -425,6 +433,11 @@ message StorageList {
   repeated nakama.api.StorageObject objects = 1;
   // Approximate total number of storage objects.
   int32 total_count = 2;
+  // Avaliable collection names in the whole of the storage
+  repeated string collections = 3;
```
mofirouz, line 437: This needs to be part of another API call to return only the collection data - so we can power the dropdown.

mofirouz: Also typo in "Avaliable"

server/console_storage.go
```diff
@@ -111,17 +123,83 @@ func (s *ConsoleServer) ListStorage(ctx context.Context, in *console.ListStorage
 		}
 		userID = &uid
 	}
-
 	var query string
+
+	collections := make([]string, 0)
+	query = "SELECT DISTINCT collection FROM storage"
```
mofirouz, line 129: I don't think we have an index on only `collection` - so we'll likely do a full index scan? Have you run a EXPLAIN statement on this?

server/console_storage.go
```diff
@@ -111,17 +123,83 @@ func (s *ConsoleServer) ListStorage(ctx context.Context, in *console.ListStorage
...
+	}
+	for rows.Next() {
+		var dbCollection string
+		if err := rows.Scan(&dbCollection); err != nil {
+			_ = rows.Close()
+			s.logger.Error("Error scanning storage collections.", zap.Any("in", in), zap.Error(err))
+			return nil, status.Error(codes.Internal, "An error occurred while trying to list storage collecctions.")
+		}
+		collections = append(collections, dbCollection)
+	}
```
mofirouz, line 143: I think there is a way to construct the query such that the result is only one row, where the value is an []TEXT (collections) - and you can load this directly into a string slice.

server/console_storage.go
```diff
@@ -111,17 +123,83 @@ func (s *ConsoleServer) ListStorage(ctx context.Context, in *console.ListStorage
...
+		} else {
+			query += " AND "
+		}
+		query += fmt.Sprintf("%s $%d", arg, i+1)
+	}
 
-	rows, err := s.db.QueryContext(ctx, query, params...)
+	sc := &consoleStorageCursor{}
+	if in.Cursor != "" {
+		cb, err := base64.RawURLEncoding.DecodeString(in.Cursor)
```
mofirouz, line 171: We usually use gob encoding for cursor - https://github.com/heroiclabs/nakama/blob/master/server/core_storage.go#L384

I suggest we stick to gob here as well.

server/console_storage.go
```diff
@@ -103,6 +114,7 @@ func (s *ConsoleServer) GetStorage(ctx context.Context, in *api.ReadStorageObjec
 }
 
 func (s *ConsoleServer) ListStorage(ctx context.Context, in *console.ListStorageRequest) (*console.StorageList, error) {
+	const limit = 5
```
mofirouz, line 117: is this correct? I think the limit should be `100`

mofirouz (changes requested):
I've left some comments, need further enhancements.

server/console_storage.go
novabyte: @michal-harish This will need to be cached on the server. The query is expensive but we also don't want an additional index because it introduces unwanted write amplification around the storage engine. Best to cache and refresh the values used in the dropdown periodically. Let me know if you want to talk this through.

server/console_storage.go
michal-harish: @novabyte do you have an idea ? a quick google / stackoverflow yields only recommendation to use row scan and append

server/console_storage.go
michal-harish: ok what we'll do is:
- create an in-memory cache set which will be nil
- when a write comes and the cache is nil do nothing
- when a call to /v2/storage-collections occurs and the cache is nil we'll do a full scan an initialize the cache (write-lock)
- when a storage object write comes and the cache is not-nil we'll update the entry for its collection value (write-lock)
- when a call to /v2/storage-collections occurs and the cache is not-nil we'll return the cache content (read-lock)

server/console_storage.go
mofirouz: https://stackoverflow.com/a/533326

The func `array_agg` exists in CockroachDB and Postgres 8+ so I think it's safe to use it.

michal-harish: so that returns a string in the format `{coll1,coll2}` so it cannot be simply scanned into a []string but instead has to be parsed into individual names and appended to the result collection  - feels a bit fiddly for otherwise simple and robust multirow scan and append. wdyt ?

console/console.proto
```diff
@@ -193,6 +193,10 @@ service Console {
     option (google.api.http).get = "/v2/console/storage";
   }
 
+  //List storage collections
+  rpc ListStorageCollections (google.protobuf.Empty) returns (StorageCollectionsList) {
+    option (google.api.http).get = "/v2/console/storage-collections";
```
mofirouz, line 198: I think this should be `/v2/console/storage/collections`

PR #507, opened 2020-12-02
Console improvements and fixes

sesposito:
Add support for deleting a leaderboard record.
Improve config storage objects file handling.


console/ui/src/app/leaderboard/records/records.component.html
```diff
@@ -20,18 +20,19 @@ <h6 class="mr-2 d-inline font-weight-bold">An error occurred: {{error}}</h6>
...
       <tr>
-        <th style="width: 315px">Owner ID</th>
+        <th style="width: 380px">Owner ID</th>
         <th>Username</th>
         <th style="width: 150px">Score</th>
         <th style="width: 150px">Subscore</th>
         <th style="width: 150px">Rank</th>
         <th style="width: 150px">No. of scores</th>
         <th style="width: 180px">Expiry Time</th>
+        <th style="width: 180px" *ngIf="deleteAllowed()">Remove</th>
```
mofirouz, line 30: Please use 90px here to match everywhere else.

console/ui/src/app/config/config.component.ts
```diff
@@ -131,6 +131,53 @@ export class ConfigComponent implements OnInit, OnDestroy {
     }
   }
 
+  // If the file is in JSON, it stringifies the value of the storage object.
+  private processFile(fIn: File, callback: (f: File) => void): void {
```
mofirouz, line 135: Whilst this is really good, I'd suggest that we make this much simpler - i.e. make the server code such that it only accepts `value` as a JSON object. This is indeed a breaking change to the behaviour of this API from Nakama2 but I think it's worth it to make it much much easier to deal with the data.

server/console_leaderboard.go
```diff
@@ -107,8 +107,8 @@ func (s *ConsoleServer) ListLeaderboardRecords(ctx context.Context, in *api.List
 			return nil, status.Error(codes.InvalidArgument, "Invalid limit - limit must be between 1 and 100.")
 		}
 		limit = in.GetLimit()
-	} else if len(in.GetOwnerIds()) == 0 || in.GetCursor() != "" {
-		limit = &wrappers.Int32Value{Value: 1}
+	} else if len(in.GetOwnerIds()) == 0 || in.GetCursor() == "" {
+		limit = &wrappers.Int32Value{Value: 10}
```
mofirouz, line 111: why this change?

server/api_leaderboard.go
```diff
@@ -114,8 +114,8 @@ func (s *ApiServer) ListLeaderboardRecords(ctx context.Context, in *api.ListLead
 			return nil, status.Error(codes.InvalidArgument, "Invalid limit - limit must be between 1 and 100.")
 		}
 		limit = in.GetLimit()
-	} else if len(in.GetOwnerIds()) == 0 || in.GetCursor() != "" {
-		limit = &wrappers.Int32Value{Value: 1}
+	} else if len(in.GetOwnerIds()) == 0 || in.GetCursor() == "" {
+		limit = &wrappers.Int32Value{Value: 10}
```
mofirouz, line 118: why this change? Should we set the limit value in the console request?

mofirouz (changes requested):
Some changes required.

server/api_leaderboard.go
sesposito: Discussed IRL

server/console_leaderboard.go
sesposito: Discussed IRL

console/ui/src/app/config/config.component.ts
sesposito: I've made the change server side instead.

mofirouz (approved):

PR #512, opened 2020-12-09
Add API Explorer endpoints to nakama console

michal-harish:


server/console.go
```diff
@@ -107,6 +111,7 @@ func StartConsoleServer(logger *zap.Logger, startupLogger *zap.Logger, db *sql.D
 		runtimeInfo:          runtimeInfo,
 		leaderboardCache:     leaderboardCache,
 		leaderboardRankCache: leaderboardRankCache,
+		api: 								  api,
```
zyro, line 114: nit: Formatting.

server/console_api_explorer.go
```diff
@@ -0,0 +1,268 @@
...
+			return nil, status.Error(codes.Internal, "Error looking up user account.")
+		}
+		callCtx = context.WithValue(context.WithValue(ctx, ctxUserIDKey{}, userUUID), ctxUsernameKey{}, dbUsername)
+	}
+	return callCtx, nil
+
+}
+
+func (s *ConsoleServer) ListApiEndpoints(ctx context.Context, _ *empty.Empty) (*console.ApiEndpointList, error) {
+	if s.methodListCache == nil {
```
zyro, line 138: This whole block is a problem for concurrency. Instead we should initialise this cache once in the server setup phase.

server/console_api_explorer.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	rpcs := make([]string, 0)
+	for _, rpc := range s.runtimeInfo.LuaRpcFunctions {
+		rpcs = append(rpcs, rpc)
+	}
+	for _, rpc := range s.runtimeInfo.GoRpcFunctions {
+		rpcs = append(rpcs, rpc)
+	}
+	for _, rpc := range s.runtimeInfo.JavaScriptRpcFunctions {
```
zyro, line 172: I can see a small problem case here: if there are multiple functions registered for the same RPC ID, in different runtimes, they can't each be invoked individually. The server picks one to prioritise - the Go runtime usually.

So returning all of them separately here might be misleading. Perhaps we should return as part of the `runtimeInfo` the "merged" list the server already has for this?

zyro: As an added bonus in the console runtime view we should mark which functions are overridden by a function registered with the same name but in a "higher priority" runtime.

server/console_api_explorer.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return &console.CallApiEndpointResponse{
+		Body:         out.Payload,
+		ErrorMessage: "",
+	}, nil
+
+}
+
+func (s *ConsoleServer) CallApiEndpoint(ctx context.Context, in *console.CallApiEndpointRequest) (*console.CallApiEndpointResponse, error) {
+	r, err := reflectMethod(s.api, in.Method)
```
zyro, line 49: I was hoping we'd have a pre-cached list of reflected methods so we can avoid doing it every single time. Is that feasible?

Also that would let us restrict which functions can be called this way. Right now a caller could try to run the `*ApiServer.Stop()` function for example - this should fail I think because of the input arguments we try to pass it but still.

zyro (changes requested):

server/console_api_explorer.go
michal-harish: Firstly @mofirouz wanted it initialized on the first call and yes, double-checking would be more appropriate here but it is a rare operation - initializing on the startup will of course solve it.

server/console_api_explorer.go
michal-harish: So this is something that will need some thinking and addition to the protos, i'll come up with something. If calling is done with a prioritization than we should look at the prirority conditions and exactly replicated them in the explorer to display the one that is going to be excuted.

server/console_api_explorer.go
michal-harish: So here again, we've agreed on the channel to only cache the list. Of course we can cache the individual methods but then it's going to be better refactored into a single cache map that serves both listing and calling. The actual call will still have to use reflection every time of course.

server/console_api_explorer.go
mofirouz: I think there is a little bit of confusion. I only wanted storage collections to be cached on first use. This is safe to cache on server start.

server/console_api_explorer.go
michal-harish: ..also filling in the values will have to be reflected, so we could create a full cache of all methods, params and protos in something like map[<method_name>]*rpcMethodReflection which would be a struct holding a reflect type of the method, reflect of the request, response and also the reflect type of the proto message of the request.

zyro (approved):

PR #513, opened 2020-12-10
add more explicit protobuf namespace for csharp

lugehorsam:
Copy of  https://github.com/heroiclabs/nakama-common/pull/11 but for our apigrpc.proto file.


novabyte:
@lugehorsam 👍

PR #515, opened 2020-12-18
Fix message defs path fields removal

sesposito:
Correctly removes message fields that are part of the path args of a request.


mofirouz (approved):
Epic 🥇

PR #519, opened 2021-01-04
Add js entrypoint config

sesposito:


novabyte (commented):
This looks good to me but it would be good for @zyro to have eyes on this as well 👀

zyro (approved):

PR #520, opened 2021-01-06
Improve js nk module httpRequest params handling

sesposito:


sesposito:
Made this commit part of another PR https://github.com/heroiclabs/nakama/pull/522.

PR #522, opened 2021-01-11
Add ReadFile runtime functions

sesposito:
Add runtime ReadFile functions to be able to read a file from a path relative to runtime.path.


novabyte:
@sesposito Nice! Looks good to me 👍

Issue #552, opened 2021-02-09
[Bug]: Error is thrown after matchmaking


MWFIAE:
## Description
Using nakama 3.1 the following code doesn't work and causes a runtime error/panic

## Steps to Reproduce
1. following coding:
```
function MatchmakerMatched(ctx: nkruntime.Context, logger: nkruntime.Logger, nk: nkruntime.Nakama, matches: nkruntime.MatchmakerResult[]) {
    logger.debug("Making Match");
    let setupState = {
        playerA: JSON.stringify(matches[0].presence),
        playerB: JSON.stringify(matches[1].presence),
    };
    const matchId = nk.matchCreate("matchhandler", setupState);
    logger.info(matchId);
    return matchId;
}

let InitModule: nkruntime.InitModule = function (ctx: nkruntime.Context, logger: nkruntime.Logger, nk: nkruntime.Nakama, initializer: nkruntime.Initializer) {
    initializer.registerMatch("matchhandler", {
        matchInit,
        matchJoin,
        matchJoinAttempt,
        matchLeave,
        matchLoop,
        matchTerminate,
    });
    initializer.registerMatchmakerMatched(MatchmakerMatched);
};
```
2. Search a match with 2 or more clients to trigger the matchmaker.

## Expected Result
Matchhandler should be called and the match should proceed normally.

## Actual Result
The MatchmakerMatched function is not called (no "Making Match" output in console). Instead following error message is thrown:
```
nakama         | panic: runtime error: invalid memory address or nil pointer dereference
nakama         | [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x17e6238]
nakama         |
nakama         | goroutine 75 [running]:
nakama         | github.com/heroiclabs/nakama/v3/server.(*RuntimeJS).invokeFunction(0xc000b19b00, 0x6, 0x1d38090, 0x11, 0x0, 0xc0018a6b40, 0x4, 0x4, 0x0, 0x0, ...)
nakama         |        github.com/heroiclabs/nakama/v3/server/runtime_javascript.go:434 +0x78
nakama         | github.com/heroiclabs/nakama/v3/server.(*RuntimeJS).InvokeFunction(0xc000b19b00, 0x6, 0x1d38090, 0x11, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
nakama         |        github.com/heroiclabs/nakama/v3/server/runtime_javascript.go:420 +0x391
[...]

PR #553, opened 2021-02-10
Fix JavaScript runtime hooks not being invoked correctly

sesposito:
Improve error handling for JS runtime function invocation.
Fix the delete button not working in the console leaderboard listing due to a grpc-gateway route override.


zyro (approved):

sesposito:
Resolves #552

PR #554, opened 2021-02-12
Fix loading registered match handlers

sesposito:


zyro (approved):

britzl:
Was this fix included in the 3.1.1 release? I seem to be running into the same thing still:

```
2021-02-17T08:14:15.000Z {"caller":"server/matchmaker.go:375","error":"Error running runtime Matchmaker Matched hook: /nakama/data/modules/main.lua:36: error creating match: not found\nstack traceback:\n\t[G]: in function 'match_create'\n\t/nakama/data/modules/main.lua:36: in main chunk\n\t[G]: ?","msg":"Error running Matchmaker Matched hook."}
```

Or is there some problem with my match module? This is the server I'm using:

    https://github.com/defold/game-xoxo-nakama-server

Can the match module require() other files like I'm doing here: https://github.com/defold/game-xoxo-nakama-server/blob/main/modules/tictactoe_match.lua#L1

zyro:
@britzl This is indeed included in the 3.1.1 release. At a glance the only problem I can see in your repo is here:
```lua
    local modulename = "tictactoe_match"
    local setupstate = { invited = matched_users }
    local matchid = nk.match_create(modulename, setupstate)
```
You're using the module name `tictactoe_match` but your Lua module for the match is called `tictactoe_state.lua` so try using module name `tictactoe_state` instead.

britzl:
> You're using the module name `tictactoe_match` but your Lua module for the match is called `tictactoe_state.lua` so try using module name `tictactoe_state` instead.

Actually not. The match logic is in modules/tictactoe_match.lua:

    https://github.com/defold/game-xoxo-nakama-server/blob/main/modules/tictactoe_match.lua

And this module requires tictactoe_state.lua. Is that the problem that a match module can't require additional files?

zyro:
I think I'm getting a bit confused about your repo layout, you have some Lua files at the top level and some in the `modules` subdirectory. Your `main.lua` is at the top level in the repo which means the match module should be `modules. tictactoe_match` (paths are important in Lua module imports!). But then based on your log message the `main.lua` file is loaded into Nakama as `modules/main.lua`.

Can you share the full startup sequence logs for the server? There should be a list of all Lua modules it has found and how they're loaded.

britzl:
Thanks for the input. I got some assistance from @novabyte to clean up the folder structure.

PR #564, opened 2021-02-26
Console user accounts are ordered by last update

RVennu:
Current ordering by ID in the console isn't really useful. Ordering by last update isn't going to cure cancer but will make browsing for specific accounts much easier.

### DIFF
--- a/server/console_account.go
+++ b/server/console_account.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"strconv"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


novabyte:
@RVennu This will perform really poorly at scale and does not meet the performance requirements we have with the Console UI's interactiveness.

RVennu:
@novabyte, I'll try one more time if that's okay. 
Is the cost of adding `CREATE INDEX users_update_time_desc_id_asc ON users (update_time DESC, id ASC)` too much for a console feature?

PR #576, opened 2021-03-25
Replace string with fmt.Sprint for int conversion

syntaqx:
Simply addresses the compiler hint typing `conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)` - Seemed like the intended use case.

`goimports` moved the internal/remote module by default, as is the coding practice for that project. If you'd like I can undo it, but I thought you'd also want that.


sesposito:
This PR has been rendered obsolete by #862.

PR #577, opened 2021-03-25
Track tool dependencies in modules

syntaqx:
Although I won't elaborate here, I've added the installation binaries for GRPC to go modules, and updated the documentation to reflect the current installation pattern (updated from their README)

- https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
- https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md
- https://github.com/grpc-ecosystem/grpc-gateway#installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,11 +142,11 @@ To build the codebase and generate all sources follow these steps.
 1. Install the toolchain.
 
    ```shell
-   go install \
-       "github.com/golang/protobuf/protoc-gen-go" \
-       "google.golang.org/grpc/cmd/protoc-gen-go-grpc" \
-       "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway" \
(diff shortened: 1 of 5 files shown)


novabyte:
Thanks for this pull request @syntaqx but we already track tools dependencies in the repo:

https://github.com/heroiclabs/nakama/blob/master/build/tools.go

syntaqx:
Aha! I must've missed that, apologies. I added it to the top of the repository because if you put it in a sub-folder and don't include that module, it's silent ignored on a `go install github.com/nakama/nakama` (or, that was the impression I got from how I've seen it used).

There is a convention where folks have put it in `tools/tools.go` or some iteration of that, but I've always seen them target the module directly when sourcing them.

Is having it in the root of the repository something you'd like me to keep this open for, or should I go ahead and close? Apologies on the duplication of efforts, still navigating the codebase!

PR #582, opened 2021-03-31
Add IAP Validation APIs

sesposito:
TODO:
- [x] Update nakama-common dependency after https://github.com/heroiclabs/nakama-common/pull/23 is merged.
- [x] Add Console listing view and API.


iap/iap.go
```diff
@@ -0,0 +1,515 @@
...
+		return ValidateReceiptAppleWithUrl(ctx, httpc, AppleReceiptValidationUrlSandbox, receipt, password)
+	}
+
+	return resp, raw, nil
+}
+
+// Validate an IAP receipt with Apple against the specified URL.
+func ValidateReceiptAppleWithUrl(ctx context.Context, httpc *http.Client, url, receipt, password string) (*ValidateReceiptAppleResponse, []byte, error) {
+	if len(url) < 1 {
+		return nil, []byte{}, errors.New("'url' must not be empty")
```
zyro, line 97: nit: Same throughout here, maybe better to use `nil` instead of `[]byte{}`.

iap/iap.go
```diff
@@ -0,0 +1,515 @@
...
+
+	if len(privateKey) < 1 {
+		return nil, nil, []byte{}, errors.New("'privateKey' must not be empty")
+	}
+
+	if len(receipt) < 1 {
+		return nil, nil, []byte{}, errors.New("'receipt' must not be empty")
+	}
+
+	token, err := GetGoogleAccessToken(ctx, httpc, clientEmail, privateKey)
```
zyro, line 289: Should a fresh access token be requested each time? If I recall last time I worked with similar Google services it should be possible to cache access tokens for some time.

iap/iap.go
```diff
@@ -0,0 +1,515 @@
...
+func ValidateReceiptHuawei(ctx context.Context, httpc *http.Client, pubKey, clientID, clientSecret, purchaseData, signature string) (*ValidateReceiptHuaweiResponse, *InAppPurchaseDataHuawei, []byte, error) {
+	if len(purchaseData) < 1 {
+		return nil, nil, []byte{}, errors.New("'purchaseData' must not be empty")
+	}
+
+	if len(signature) < 1 {
+		return nil, nil, []byte{}, errors.New("'signature' must not be empty")
+	}
+
+	data := &InAppPurchaseDataHuawei{PurchaseType: 2} // Set to placeholder value because field is omitted by prod purchase.
```
zyro, line 431: As discussed use a sentinel/placeholder value outside any possible valid range. Perhaps `-1` would be best?

zyro (changes requested):

console/console.swagger.json
```diff
@@ -9,11 +9,6 @@
       "email": "hello@heroiclabs.com"
     }
   },
-  "tags": [
```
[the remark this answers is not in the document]
sesposito: I've run `go generate -x` in the console as usual and this is removed, should I revert it?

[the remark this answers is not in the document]
zyro: It's interesting that each operation still contains a `"tags"` field, but the global one looks like it's been removed. I think if our codegens don't need the top-level tags then leave it removed.

iap/iap.go
```diff
@@ -202,21 +207,44 @@ func DecodeReceiptGoogle(receipt string) (*ReceiptGoogle, error) {
...
+
 // Request an authenticated context (token) from Google for the Android publisher service.
-func GetGoogleAccessToken(ctx context.Context, httpc *http.Client, email string, privateKey string) (string, error) {
+// https://developers.google.com/identity/protocols/oauth2#serviceaccount
+func getGoogleAccessToken(ctx context.Context, httpc *http.Client, email string, privateKey string) (string, error) {
 	const authUrl = "https://accounts.google.com/o/oauth2/token"
 
+	cachedTokenGoogle.RLock()
+	if cachedTokenGoogle.AccessToken != "" && !cachedTokenGoogle.Expired() {
+		return cachedTokenGoogle.AccessToken, nil
```
zyro, line 231: Needs `cachedTokenGoogle.RUnlock()` before the return here.

iap/iap.go
```diff
@@ -381,11 +408,25 @@ type AccessTokenHuawei struct {
...
+func (at *accessTokenHuawei) Expired() bool {
+	return at.ExpiredAt-accessTokenExpiresGracePeriod <= time.Now().Unix()
+}
+
+func getHuaweiAccessToken(ctx context.Context, httpc *http.Client, clientID, clientSecret string) (string, error) {
 	const authUrl = "https://oauth-login.cloud.huawei.com/oauth2/v3/token"
 
+	cachedTokenHuawei.RLock()
+	if cachedTokenHuawei.AccessToken != "" && !cachedTokenHuawei.Expired() {
+		return cachedTokenHuawei.AccessToken, nil
```
zyro, line 427: Needs `cachedTokenHuawei.RUnlock()` before the return here.

server/core_purchase.go
```diff
@@ -410,11 +406,15 @@ returning transaction_id, create_time, update_time"
 	if err != nil {
 		return nil, err
 	}
+	defer rows.Close()
```
zyro, line 409: There's already a `rows.Close()` after the loop so I would just move this into the loop ahead of the error return, rather than as a `defer` here.

zyro (changes requested):

zyro (approved):

PR #585, opened 2021-04-07
Fix js fn incorrectly returned timestamps

sesposito:


novabyte:
Looks good to me @sesposito just needs a CHANGELOG entry 👍

PR #592, opened 2021-04-16
Fix console purchase scoping and pagination issues

sesposito:


novabyte:
Looks good to me @sesposito 👍

Issue #595, opened 2021-04-19
Nakama server 3.2.1 doesn't start


sadegh-askari:
Hi
I download last nakama server (3.2.1) for windows
when i try to start server in cmd, i got this error:
{"level":"fatal","ts":"2021-04-20T02:08:49.858+0430","caller":"migrate/migrate.go:91","msg":"Could not find migrations","error":"open sql\\20180103142001_initial_schema.sql: file does not exist"}

novabyte:
@sadegh-askari This could be an issue with how the latest release was packaged. Is it possible you could try the 3.1.1 release of the server and report back?

sadegh-askari:
Yes
I downloaded version 3.1.1 And it works properly

novabyte:
@sadegh-askari Thanks. I suspect the issue is with the new Go 1.16 embed package we switched over to use and how it handles filesystem paths inside the virtual filesystem on Windows. We'll take a look and patch it as part of the next server release.

sadegh-askari:
TNX
I forked server and fix this problem for myself.
but same issue exist when i want to open console at localhost:7351
{"level":"error","ts":"2021-04-20T04:25:01.935+0430","caller":"server/console.go:290","msg":"Failed to open index file.","error":"open ui\\dist\\index.html: file does not exist"}

michal-harish:
the fix is ready and tested on windows and osx

michal-harish:
@sadegh-askari thanks for raising this. The fix is now merged and will be part of the next release.

PR #597, opened 2021-04-20
#595 use path.Join for go embed filesystems instead of filepath.Join

michal-harish:
…because embed fs always uses forward slash separators


novabyte (approved):
@michal-harish Looks good to me 👍

PR #612, opened 2021-05-09
Store email and DisplayName if social provider has one.

mofirouz:
Store email (update only if one not set) and display name (how GoogleAuth/Link already works) for Facebook and Apple social providers.

### DIFF
--- a/server/core_authenticate.go
+++ b/server/core_authenticate.go
@@ -20,10 +20,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"google.golang.org/protobuf/types/known/timestamppb"
 	"strconv"
 	"strings"
 	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/gofrs/uuid"
 	"github.com/heroiclabs/nakama-common/api"
@@ -77,8 +78,8 @@ func AuthenticateApple(ctx context.Context, logger *zap.Logger, db *sql.DB, clie
 
 	// Create a new account.
 	userID := uuid.Must(uuid.NewV4()).String()
-	query = "INSERT INTO users (id, username, apple_id, create_time, update_time) VALUES ($1, $2, $3, now(), now())"
-	result, err := db.ExecContext(ctx, query, userID, username, profile.ID)
+	query = "INSERT INTO users (id, username, email, apple_id, create_time, update_time) VALUES ($1, $2, $3, $4, now(), now())"
+	result, err := db.ExecContext(ctx, query, userID, username, profile.Email, profile.ID)
 	if err != nil {
 		if e, ok := err.(pgx.PgError); ok && e.Code == dbErrorUniqueViolation {
 			if strings.Contains(e.Message, "users_username_key") {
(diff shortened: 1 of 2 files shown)


server/core_authenticate.go
```diff
@@ -670,8 +671,8 @@ func AuthenticateGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, cli
 
 	// Create a new account.
 	userID := uuid.Must(uuid.NewV4()).String()
-	query = "INSERT INTO users (id, username, google_id, display_name, avatar_url, create_time, update_time) VALUES ($1, $2, $3, $4, $5, now(), now())"
-	result, err := db.ExecContext(ctx, query, userID, username, googleProfile.Sub, displayName, avatarURL)
+	query = "INSERT INTO users (id, username, email, google_id, display_name, avatar_url, create_time, update_time) VALUES ($1, $2, $3, $4, $5, now(), now())"
```
novabyte, line 674: Same problem as above.

server/core_authenticate.go
```diff
@@ -431,8 +432,8 @@ func AuthenticateFacebook(ctx context.Context, logger *zap.Logger, db *sql.DB, c
 
 	// Create a new account.
 	userID := uuid.Must(uuid.NewV4()).String()
-	query = "INSERT INTO users (id, username, facebook_id, create_time, update_time) VALUES ($1, $2, $3, now(), now())"
-	result, err := db.ExecContext(ctx, query, userID, username, facebookProfile.ID)
+	query = "INSERT INTO users (id, username, display_name, email, facebook_id, create_time, update_time) VALUES ($1, $2, $3, $4, now(), now())"
```
novabyte, line 435: There aren't enough placeholder parameters to match the expected inputs in this SQL.

server/core_link.go
```diff
@@ -44,15 +45,15 @@ func LinkApple(ctx context.Context, logger *zap.Logger, db *sql.DB, config Confi
 	}
 
 	res, err := db.ExecContext(ctx, `
-UPDATE users
-SET apple_id = $2, update_time = now()
+UPDATE users AS u
+SET apple_id = $2, email = COALESCE(u.email, $3), update_time = now()
```
novabyte, line 49: I would suggest we adjust part of that query to be: `COALESCE(NULLIF(u.email, ''), $3)`. This way we can set the `$3` as the email address even if the existing email field is empty rather than NULL.

server/core_link.go
```diff
@@ -202,15 +203,15 @@ func LinkFacebook(ctx context.Context, logger *zap.Logger, db *sql.DB, socialCli
 	}
 
 	res, err := db.ExecContext(ctx, `
-UPDATE users
-SET facebook_id = $2, update_time = now()
+UPDATE users AS u
+SET facebook_id = $2, display_name = $3, email = COALESCE(u.email, $4), update_time = now()
```
novabyte, line 207: Same as above.

server/core_link.go
```diff
@@ -325,15 +326,15 @@ func LinkGoogle(ctx context.Context, logger *zap.Logger, db *sql.DB, socialClien
 	}
 
 	res, err := db.ExecContext(ctx, `
-UPDATE users
-SET google_id = $2, display_name = $3, avatar_url = $4, update_time = now()
+UPDATE users AS u
+SET google_id = $2, display_name = $3, avatar_url = $4, email = COALESCE(u.email, $5), update_time = now()
```
novabyte, line 330: Same as above.

novabyte (approved):

PR #624, opened 2021-06-03
add empty bodies to post methods in apigrpc.proto

lugehorsam:
This is a slight workaround for https://github.com/grpc-ecosystem/grpc-gateway/issues/2093 so that we don't need to manually patch apigrpc.swagger.json after generating it.

One other thing I noticed -- it seems like the `protoc` invocation when running `go generate...` is dependent on my local protobuf installation. This was unexpected given the tool dependencies listed in https://github.com/heroiclabs/nakama/blob/master/build/tools.go.

Should we look at including the `protoc` tool itself in the tool dependencies?

### DIFF
--- a/apigrpc/apigrpc.proto
+++ b/apigrpc/apigrpc.proto
@@ -529,7 +529,8 @@ service Nakama {
   // Promote a set of users in a group to the next role up.
   rpc PromoteGroupUsers (api.PromoteGroupUsersRequest) returns (google.protobuf.Empty) {
     option (google.api.http) = {
-      post: "/v2/group/{group_id}/promote"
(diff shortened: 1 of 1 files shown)


novabyte:
@lugehorsam You cannot pin the protoc compiler itself via a tools dependency as far as I'm aware. We've always listed it as an external tool that can be installed via a system package manager. It's only the protoc plugins that we use for Go code generation that can be pinned (which we've done).

At one point I looked at Bazel to do our builds which can be instructed to download prerequisite parts of the build toolchain itself but the Go support in Bazel was quite immature at the time. I think we just update what we mention as prerequisites to develop on the codebase in our README.

zyro (approved):

PR #626, opened 2021-06-09
Remove analytics

feliperuhland:
The console was built with segment enable by default with two different
profiles: "dev" and "production". The problem with that approach is that
Heroic Labs received telemetry data from every console used as an
open-source platform.

Two platforms (`segment.io` and `heapanalytics.com`) were collecting
telemetry data from Nakama's console.

This commit aims to change that behavior and does not start telemetry by
default but allows Heroic Labs to enable by command line.

From now on, `ng serve` and `ng build` will not start telemetry by
default

Topic: https://forum.heroiclabs.com/t/opt-out-analytics-data/1692/2


novabyte:
@feliperuhland Thanks for the pull request but this does not solve the issue you had which was discussed on the forums. There's already a clear way to disable the analytics (`env NAKAMA_TELEMETRY=false nakama ....`) though I think it was agreed that it's a bug where it's not disabled in the Nakama Console. We'll resolve this ourselves and it will be part of the next server release.

I will leave this open until that release is made. Thanks

feliperuhland:
Hi @novabyte, thanks for your answer. Maybe I wasn't clear enough about the analytics so let me explain again.

I want to run Nakama as a game server for a small casual game, and I downloaded the binary and started developing the game.
Nakama is a great project and allows the game developer to focus on the game logic instead of worry about backend subjects.
However, I noticed that my DNS filter blocked many requests to `segment.io` and `heapanaytics.com`, and after some investigation, I realized the Nakama Console was making those requests. I assume that it was a _simple mistake that Heroic Labs_ forgot to disable analytics before compiling the binary, so I clone the project to understand what was happening.

Meanwhile, I wrote a [forum message](https://forum.heroiclabs.com/t/opt-out-analytics-data/1692) asking to disable those analytics requests. The answer was to set an environment variable `NAKAMA_TELEMETRY` to disable analytics. In fact, there was **another** analytics there, but this time it was sending by the backend. The answer was right, the environment variable disables that backend analytics, but the one I was complaining about was executing on the frontend.

My surprise was to find two differents `segment.io` keys:

- [one](https://github.com/heroiclabs/nakama/blob/master/console/ui/src/environments/environment.ts#L17) for development running `ng serve` and;
- [another one](https://github.com/heroiclabs/nakama/blob/master/console/ui/src/environments/environment.prod.ts#L17) for production environment after build running `npm run-scripts build`.

Again, I thought it could be a _simple mistake that Heroic Labs_ forgot to remove the keys from the open-source repository.

That said, I opened this pull request to invert the logic and remove analytics from Nakama Console by default and allow Heroic Labs developers to enable whenever they need it: for [development](https://github.com/heroiclabs/nakama/pull/626/files#diff-450711a7e63ceea9af0834da7ab95229ebfb3d1a1bb91c2531605fb03bf6ff83R11) or [production](https://github.com/heroiclabs/nakama/pull/626/files#diff-450711a7e63ceea9af0834da7ab95229ebfb3d1a1bb91c2531605fb03bf6ff83R27).

Also, it would be better to have binaries and Docker images without analytics libraries because game developers don't want to be tracked on their dev/prod environments.

Lastly, this pull request will break the Heroic Labs pipeline - sorry about that; I'm sure that will be fixed right away by adding the flags I mentioned on the readme. Also, I can improve the README to describe better how to enable the analytics if you want it.

Thanks again for your time.

Have a nice day :)

zyro:
Addressed cleanly in https://github.com/heroiclabs/nakama/commit/f07fce46a91b0ef1677ac8d48828d1e6faaf2dc4, this PR is now redundant. Thanks for raising this! 👍

PR #635, opened 2021-06-25
Add console account auth view device id copy-to-clipboard button

### LINKED ISSUE #630: Nakama Console: Add a way to copy the Device Id
It would be nice to be able to copy the device id from the player account.

Thanks

sesposito:
Resolves #630


mofirouz (approved):
Nice one @sesposito

PR #651, opened 2021-07-25
Add GitHub Actions CI

slonopotamus:
Notes:

~This PR doesn't use `xgo` suggested by [build/README.md](https://github.com/heroiclabs/nakama/blob/master/build/README.md) for several reasons:~

~1. `docker pull techknowlogick/xgo:latest` is [slow](https://github.com/slonopotamus/nakama/runs/3155601068?check_suite_focus=true) (almost 2 mins)~
~2. I don't see what additional value `xgo` has over `GOOS`/`GOARCH`~
~3. It [fails to build](https://github.com/slonopotamus/nakama/runs/3155601068?check_suite_focus=true#step:6:109). If you think I didn't follow `build/README.md` instructions properly, please tell we where I'm wrong.~

Now actually it *does* use `xgo`.

Also, this PR doesn't do [Full Source Build](https://github.com/heroiclabs/nakama#full-source-builds). Required steps can be added easily, but I'd like to move in small steps.

Also, this PR doesn't run tests. I failed to find instructions how to run tests in nakama repo anywhere. `go test` says:
```
$ go test
?       github.com/heroiclabs/nakama/v3 [no test files]
```

And lastly, badge in README will be 404 until this PR is merged. You don't need to perform any manual actions in order to activate GitHub Actions, they will start working automagically. For now, you can see example build runs in [my fork](https://github.com/slonopotamus/nakama/actions/runs/1065025672).

In the future, it is possible to automate release process (including Docker images and publishing of tarballs to GitHub Releases).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,39 @@
+name: CI
+on: [push, pull_request]
+
+env:
+  GOVERSION: 1.16.5
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        include:
+          - arch: arm64
+            os: linux
+          - arch: amd64
+            os: linux
+          - arch: amd64
+            os: darwin
+          - arch: amd64
+            os: windows
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GOVERSION }}
+      - run: env GO111MODULE=off go get -u src.techknowlogick.com/xgo
+      - run: |
+          ~/go/bin/xgo \
+          -go go-${{ env.GOVERSION }} \
+          --targets=${{ matrix.os }}/${{ matrix.arch }} \
+          --trimpath \
+          --ldflags "-s -w -X main.version=$(git describe --tags) -X main.commitID=$(git rev-parse --short HEAD 2>/dev/null)" \
+          .
+      - uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.os }}-${{ matrix.arch }}
+          path: github.com/heroiclabs/nakama/*
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+          - amd64
+          - arm64
+        os:
+          - linux
+        include:
+          - arch: amd64
+            os: darwin
+          - arch: amd64
+            os: windows
+    runs-on: ubuntu-20.04
```
novabyte, line 18: We should use a Debian base image if we can because that's what the container releases use.

novabyte (commented):
@slonopotamus Thanks for the pull request 🙏 

> I don't see what additional value xgo has over GOOS/GOARCH

The value is that it allows you to do a cross-compiler build with CGO enabled which is required for Go plugins to work. We use Go plugins to allow developers to write and compile their Go code with the `--buildmode=plugin` option of the compiler to generate a shared object. This shared object can be loaded by the game server with the linker as part of the Go runtime.

> It fails to build. If you think I didn't follow build/README.md instructions properly, please tell we where I'm wrong.

There's a regression in the `go get` sub-command of the toolchain which we've not had the chance to address yet. It over scans older versions of the dependency tree and ends up with an error. We have a version of xgo we've used so far as a workaround.

> Also, this PR doesn't do Full Source Build. Required steps can be added easily, but I'd like to move in small steps.

Agreed. I think we can avoid a full source build for now and we can consider how best to utilize that approach at a later stage. At the moment we check-in the generated source files from the protobuf toolchain anyway to ensure that we have repeatable builds.

> Also, this PR doesn't run tests. I failed to find instructions how to run tests in nakama repo anywhere. go test says:

We have a small test suite in the main source tree but the majority of the test suite is not public. We have customer orientated tests we've amassed over time and are unable to make those public because it would break our NDAs. At some point we'll split the test suite up more cleanly into public and internal/private codebases.

I've left a small comment but will merge it after the change 👍

slonopotamus:
> We have a small test suite in the main source tree

Well, no matter how big it is, it would be great if it was run as part of CI build.

.github/workflows/ci.yml
slonopotamus: This is the list of available base images: https://github.com/actions/virtual-environments#available-environments
Note that this is not a Docker container, this is a full-blown virtual machine.

What we can do is to perform build inside docker container. Stay tuned, I'm updating this PR.

slonopotamus:
> There's a regression in the go get sub-command of the toolchain which we've not had the chance to address yet.

Possibly you know in what version in worked in the past and we can use `xgo -go <go version>` to use that particular version until the issue is fixed?

slonopotamus:
Okay, I've managed to use `xgo`, see updated changes. And here's example build run: https://github.com/slonopotamus/nakama/actions/runs/1067456126

.github/workflows/ci.yml
slonopotamus: Done, it uses `xgo` now.

slonopotamus:
Some kind of GitHub Actions config was merged in #1048.

PR #665, opened 2021-08-14
Fix wrong http_key parameter

zJudGenie:
As already [said on Gitter](https://gitter.im/heroiclabs/nakama?at=611574639484630efa3556bc), this fixes the problem of some SDKs generating code based on the apigrpc.swagger.json file, like the [nakama-godot](https://github.com/heroiclabs/nakama-godot/blob/master/addons/com.heroiclabs.nakama/api/NakamaAPI.gd#L5172) addon

### DIFF
--- a/apigrpc/apigrpc.proto
+++ b/apigrpc/apigrpc.proto
@@ -68,8 +68,8 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
       key: "HttpKeyAuth";
       value: {
         type: TYPE_API_KEY;
-        in: IN_HEADER;
-        name: "http_key";
+        in: IN_QUERY;
+        name: "httpKey";
       }
     }
   }
--- a/apigrpc/apigrpc.swagger.json
+++ b/apigrpc/apigrpc.swagger.json
@@ -4802,8 +4802,8 @@
     },
     "HttpKeyAuth": {
       "type": "apiKey",
-      "name": "http_key",
-      "in": "header"
+      "name": "httpKey",
+      "in": "query"
     }
   },
   "security": [
--- a/server/api.go
+++ b/server/api.go
@@ -149,7 +149,7 @@ func StartApiServer(logger *zap.Logger, startupLogger *zap.Logger, db *sql.DB, p
 			q := r.URL.Query()
 			p := make(map[string][]string, len(q))
 			for k, vs := range q {
-				if k == "http_key" {
+				if k == "httpKey" {
 					// Skip Nakama's own query params, only process custom ones.
 					continue
 				}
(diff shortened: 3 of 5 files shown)


zJudGenie:
Maybe this needs to be changed as well https://github.com/heroiclabs/nakama-common/blob/e507376361e7cacd2fbd4a0a04e65ae9c67a53db/api/api.pb.go#L5677

jasonwinterpixel:
Does this Pull Request assume that the nakama godot sdk's rpc_async_with_key doesn't work? The nakama godot sdk works as is using 'http_key'. This PR looks like it breaks compatability by renaming http_key to httpKey for no reason.

zJudGenie:
> Does this Pull Request assume that the nakama godot sdk's rpc_async_with_key doesn't work? The nakama godot sdk works as is using 'http_key'. This PR looks like it breaks compatability by renaming http_key to httpKey for no reason.

Hi, are you sure that Nakama Godot SDK works as is ? I used it some time ago(when I created the PR) and I had to manually change the param [here](https://github.com/heroiclabs/nakama-godot/blob/master/addons/com.heroiclabs.nakama/api/NakamaAPI.gd#L5200), because the Nakama Godot source is generated by this [codegen ](https://github.com/heroiclabs/nakama-godot/tree/master/codegen) that takes as input the Swagger documentation of the Nakama API, and as you can see [here](https://github.com/heroiclabs/nakama/blob/master/apigrpc/apigrpc.swagger.json#L2552) the param is called `httpKey` as in the Godot SDK, but the Nakama[ code behind the RPC HTTP API](https://github.com/heroiclabs/nakama/blob/master/server/api_rpc.go#L67) try to search the param `http_key` giving me `StatusUnauthorized`

Of course it breaks compatibility, so maybe it is better to edit the Swagger API and rename `httpKey` in `http_key`

jasonwinterpixel:
You are right. My apologies.

Nakama-godot is broken on master. We are using an older version of nakama-godot that uses "http_key" properly. Nice find.

Though I recommend you dont change all the references of of http_key to httpKey. Rather I recommend that we fix nakama-godot to use the proper query parameter.

This is the commit that broke nakama-godot:

https://github.com/heroiclabs/nakama-godot/commit/b6272c4547c3d4d54a7e485c2d70eaa167969c08#diff-741624c7e9958bec7aed7aa872311130d26741a458ac63dca4ec1f42d6346e7aL3988

jasonwinterpixel:
I think this is the correct fix:
https://github.com/heroiclabs/nakama/pull/787

zyro:
@zJudGenie This has been a longstanding issue we've had to clarify, and we've settled on `snake_case` for all request and query parameter fields. The console API side has already been updated https://github.com/heroiclabs/nakama/commit/a86e842118d84129fccfadc260412660cca94fe1 and the client API side will follow shortly.

To be clear, the server has always intended to process `snake_case` for parameters (and as far as we're aware that works correctly). It's only the swagger spec and corresponding code generators that are being updated to tighten up these inconsistencies.

PR #671, opened 2021-08-18
Improve Go runtime match init validation

sesposito:


zyro (approved):
Please update the changelog as well.

zyro:
After discussion with @sesposito we'll park this for now and use an alternative solution.

PR #672, opened 2021-08-18
Added Cloud Build script and related changes in the aws-marketplace

michal-harish:
I've also simplified the number of CF template parameters by using the stack name for the derived resource names and dropping the Region parameter and using the one user is targeting to deploy to.


mofirouz (approved):
Nice one 👍

PR #680, opened 2021-08-27
Adjusted default value of updateLedger of walletsUpdate

LucasGaspar:
On the documentation and on the code tooltips it says that the default is `true`, while on practice the default was `false`, changed the default to `true`.

image

image


zyro:
@LucasGaspar Thanks for the PR but in this case the documentation is wrong, the default for this flag should be `false`. We'll address this in PR #681.

PR #682, opened 2021-08-28
Configure whether or not to add the Lua stacktrace in the API response

zJudGenie:
Add a parameter to the config to choose whether or not to add the Lua stacktrace in the API response, instead of checking the log level. For example we want to use the DEBUG log level but we don't want to leak the stacktrace to someone trying to do bad things.


novabyte:
@zJudGenie This pull request looks good but will break backwards compatibility so we'll need to see if the field has been set in the config as an override and apply it only then if possible. 

I also have a small question for you about your operational environment for the game server; why would you run the server in production at DEBUG level logs?

zJudGenie:
@novabyte I don't have a special answer to this question, simply we are a small team working on an open source FPS and we only have one game server, so I prefer to log everything to notice any strange activity.

To avoid breaking compatibility I will add another commit to this PR

zyro (approved):

Issue #693, opened 2021-10-01
invalid memory address or nil pointer dereference - lua pcall


kavenGw:
## Description
lua pcall over but luastate is Closed caused by (match stop with call queue is full)

## Steps to Reproduce
1. pcall(nk.storage_write,xxx) in match_loop
2. a lot of presences join match to trigger call queue is full

## Actual Result
server crash

kavenGw:
move to 694

Issue #636, opened 2021-07-04
Message read error in one match closing all server authoritative matches. 


ragavendranbala:
I'm facing issues like 

"websocket: close 1006 (abnormal closure): unexpected EOF" 

"read tcp 172.20.0.3:7350->106.206.240.93:44654: i/o timeout" (Attached Log files)
Say there are 5 matches running on my server and if one of the match faced these type of issue. it closing all the matches and which are not affected (other 4 matches).

I have raised the issue in forum and waiting for update for a couple of weeks. 

I have attached Docker compose file and Config file. Kindly check this

[Docker Copmpose and config.zip](https://github.com/heroiclabs/nakama/files/6759517/Docker.Copmpose.and.config.zip)

[logfile.log](https://github.com/heroiclabs/nakama/files/6759501/logfile.log)

Ragavendran-DX:
More info : 
Unity : 2020.3.12
Nakama Unity Version : 2.9.5
Server version: 20a55b5f+20a55b5f

And Same issue on [v3.4.0]

sesposito:
@ragavendranbala I've taken a look at the provided files and the only thing I can infer is that the client seems to be terminating the connection unexepectedly, this could be just bad logic on the client side.

I don't see how the error you're describing would terminate multiple matches uness there's some issue with your match handler code.

Either way I cannot reproduce, so I'll be closing this issue for the time being.

PR #694, opened 2021-10-01
invalid memory address or nil pointer dereference - lua pcall. (#693)

kavenGw:


kavenGw:
Description
lua pcall over but luastate is Closed caused by (match stop with call queue is full)

Steps to Reproduce
pcall(nk.storage_write,xxx) in match_loop
a lot of presences join match to trigger call queue is full
Actual Result
server crash

ragavendranbala:
This pull request looks like fix my issue :
https://github.com/heroiclabs/nakama-unity/issues/114
https://github.com/heroiclabs/nakama/issues/636

@kavenGw can you tell me how to build a docker file from this pull request? Thanks in advance

kavenGw:
> This pull request looks like fix my issue : [heroiclabs/nakama-unity#114](https://github.com/heroiclabs/nakama-unity/issues/114) #636
> 
> @kavenGw can you tell me how to build a docker file from this pull request? Thanks in advance

I just read your issue, but I don't think we have the same reason. I'm using Nakama Lua runtime and my issue is caused by one of my matches being stopped by the call queue is full. But there is a DB read in match_loop, the DB read will return failed to the Lua runtime(which the Lua state is already disposed when the match stopped).

Anyway, if you want to try my code. Just use the default docker file. I just change 2 lines.

sesposito:
@kavenGw do you have more info or a stacktrace you could provide on the panic you've observed?

zyro:
Addressed cleanly in 6d8195b99d4c59247e0aa98b92f1c8aa806be19d, stack should only have been `nil` when all match function executions ended.

PR #714, opened 2021-11-16
refactor matchmaker process

mschoch:
This refactor attempts to split the Matchmaker process method up into smaller methods of related functionality.  My understanding of the logic is roughly:

- Iterate over active items in random order
- Build a Bluge search which finds all other items which match what this item is looking for
- Execute the search
- Skip over some Bluge results at this point
    - Matches itself
    - Matches other people already in party, etc
- Each match now either becomes part of an existing combination, or the start of a new combination
- This resulting combination is given a final check as follows:
    - Final size equals max count or this is the last iteration and the final size is > min count
    - Resulting final size satisfies each members desired min count
- If a match has been found, the internal state is updated such that subsequent matches cannot be made against the already matched items.
- Main loop continues to find/build matches for subsequent active items

This refactor is still not perfect, but it brings the main process function under 100 lines.  Note, the helper methods which operate with the lock held have the suffix LOCKED.

Whether or no we want to merge this refactor is a separate issue, it was mostly a way for me to try to test my understanding of the code.

On the topic of ensuring mutual matches, it would seem that the correct place to add this check would be in "skip over some bluge results" phase.  At this point, if we had an efficient way to check that the search hit's query also matches the outer for-loop's hit, we could detect and eliminate non-mutual matches at this phase.  And to my understanding, the rest of the logic would work correctly unchanged.

At the moment I have 2 ideas:
1.  Run another Bluge query on the same snapshot as the original query.  Obviously this immediately raises concerns about some sort of n<sup>2</sup> query execution, but we could do this with an additional MUST clause on the outer for-loop's `_id`.  This might be sufficient to keep the cost of the query down to an acceptable level.  This should be easy to prototype, but may require additional testing to validate the performance implications.
2. Use something like [bleve sear](github.com/blevesearch/sear) to allow for fast single-document matches a query execution.  This might yield a leaner/faster solution, but would also involve additional effort as sear would have to be adapted to work alongside bluge.

Another unrelated observation while in this section of code: the bluge request today uses top-N with `n=len(activeIndexes)`, which generally should be replaced with an AllMatches request.  This should work for Nakama as we process each search hit one at a time anyway, and often we may find what we're looking for early anyway.  By doing this, we would avoid the need to allocate as many large slices for the entire result set, because we process them as we go.

### DIFF
--- a/server/matchmaker.go
+++ b/server/matchmaker.go
@@ -202,133 +202,30 @@ func (m *LocalMatchmaker) process(batch *index.Batch) {
 			delete(m.activeIndexes, ticket)
 		}
 
-		indexQuery := bluge.NewBooleanQuery()
-		// Results must match the query string.
-		parsedIndexQuery, err := ParseQueryString(index.Query)
+		searchRequest, err := m.buildSearchRequestLOCKED(index)
 		if err != nil {
 			m.logger.Error("error parsing query string", zap.Error(err))
 			continue
 		}
-		indexQuery.AddMust(parsedIndexQuery)
 
-		// Results must also have compatible min/max ranges, for example 2-4 must not match with 6-8.
-		minCountRange := bluge.NewNumericRangeInclusiveQuery(
-			float64(index.MinCount), math.Inf(1), true, true).
-			SetField("min_count")
-		indexQuery.AddMust(minCountRange)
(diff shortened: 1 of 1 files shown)


zyro:
After discussion with @mschoch we've decided to park this until further improvements land in the matchmaker. We can revisit refactoring then. 👍

PR #715, opened 2021-11-16
fix behavior with query string *

mschoch:
A query string of * should return all matches, however this
behavior was lost during the Bleve->Bluge conversion.  Additional
unit tests have been added to ensure correct behavior going
forward.


mschoch:
@zyro both of these tests were failing prior to this fix, so depending on how widely this feature is used, this may be urgent.

zyro (approved):

PR #717, opened 2021-11-22
Improve js runtime custom error rpc response payload

### LINKED ISSUE #689: Allow setting HTTP Response code for webhooks
I am currently integrating with a service that requires the following behaviour from my webhook:

1) On success a particular JSON body. (This is already 100% doable with Nakama)
2) On error a particular HTTP response code AND a specific JSON body, in this case a code and message field: 

For instance:
```
{"code":"INVALID_PARAMETER","message":"Failed to parse request"}
```

However, this is currently not possible because the only way to set an HTTP response with Nakama is to actually throw an error and once an error is thrown Nakama wraps your JSON as follows:

```
{"code":3,"error":{"type":"uncaughtExceptionError"},"message":"{\"code\":\"INVALID_PARAMETER\",\"message\":\"Failed to parse request\"}"}
```

This causes my S2S call to fail.

-----------------------

Ideally, Nakama should allow setting the HTTP response code directly. Perhaps through the context variable? This can be restricted to just webhooks if need be. Another possibility is that we provide an additional field when we throw an error that instructs Nakama not to wrap the output, perhaps:

```
  let xerror: CustomError = { code: "INVALID_PARAMETER", message: message };
  let error: nkruntime.Error = { message: JSON.stringify(xerror), code: code, doWrap: false };
```

If Nakama sees the doWrap false flag, it could just return the message raw...

sesposito:
If 'message' field is set and non-empty in the thrown error object, set it in the returned json payload response.

Resolves #689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ The format is based on [keep a changelog](http://keepachangelog.com) and this pr
 - Fix handling of leaderboard record writes that do not need to update the database.
 - Fix parsing edge case in TypeScript/JavaScript runtime storage delete operations.
 
+### Changed
+- Set JS runtime custom error message as the returned payload message in RPC requests.
+
 ## [3.9.0] - 2021-10-29
 ### Added
 - Allow creation of relayed matches with a name. Names will be mapped to match identifiers.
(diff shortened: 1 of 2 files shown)


zyro (approved):
Don't forget to update the changelog.

PR #720, opened 2021-11-24
update match tests to not rely on time.Sleep

mschoch:
The purpose of this commit is to improve the match related tests
to manually drive the update/processing logic, and not rely on
time.Sleep to allow for "some" processing to occur.

To do this, the MatchRegistry and Matchmaker are updated to
allow for a config value of 0 for the label update interval
and for the matchmaking interval.  When configured with such
a value, they will not automatically process data, instead
they rely on manual invocation of their process methods.

Finally, the tests are updated to configure these components with
the 0 value, and then in test logic, sleep statements are replaced
with the appropriate call to invoke the processing logic once.

### DIFF
--- a/server/match_registry.go
+++ b/server/match_registry.go
@@ -177,21 +177,19 @@ func NewLocalMatchRegistry(logger, startupLogger *zap.Logger, config Config, ses
 		stoppedCh: make(chan struct{}, 2),
 	}
 
-	if config.GetMatch().LabelUpdateIntervalMs > 0 {
-		go func() {
-			ticker := time.NewTicker(time.Duration(config.GetMatch().LabelUpdateIntervalMs) * time.Millisecond)
-			batch := bluge.NewBatch()
-			for {
-				select {
-				case <-ctx.Done():
-					ticker.Stop()
-					return
-				case <-ticker.C:
-					r.processLabelUpdates(batch)
-				}
+	go func() {
+		ticker := time.NewTicker(time.Duration(config.GetMatch().LabelUpdateIntervalMs) * time.Millisecond)
+		batch := bluge.NewBatch()
+		for {
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
+				return
+			case <-ticker.C:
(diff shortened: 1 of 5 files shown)


server/matchmaker.go
```diff
@@ -161,18 +161,20 @@ func NewLocalMatchmaker(logger, startupLogger *zap.Logger, config Config, router
 		activeIndexes:  make(map[string]*MatchmakerIndex),
 	}
 
-	go func() {
```
zyro, line 164: `config.GetMatchmaker().IntervalSec` values <= 0 are not valid, so I don't think we should add this workaround. For test purposes I like calling the `processLabelUpdate` function directly in the test case, but leave the goroutine in place and just set the config `IntervalSec` to a very high value in the test cases.

server/match_registry.go
```diff
@@ -177,19 +177,21 @@ func NewLocalMatchRegistry(logger, startupLogger *zap.Logger, config Config, ses
...
-		ticker := time.NewTicker(time.Duration(config.GetMatch().LabelUpdateIntervalMs) * time.Millisecond)
-		batch := bluge.NewBatch()
-		for {
-			select {
-			case <-ctx.Done():
-				ticker.Stop()
-				return
-			case <-ticker.C:
-				r.processLabelUpdates(batch)
+	if config.GetMatch().LabelUpdateIntervalMs > 0 {
```
zyro, line 180: Same as the matchmaker, we should always start the goroutine.

zyro (changes requested):

server/matchmaker.go
mschoch: Yes I agree, that is a good suggestion.  It returns the main code to the original implementation, and now all changes are confined to the tests.

mschoch: The change has been made in the latest commit.

server/match_registry.go
mschoch: Done as well.

zyro (approved):

PR #733, opened 2021-12-07
Correctly handle binary data in js runtime matches

### LINKED ISSUE #659: Server authoritative bytes data gets corrupted on TypeScript
## Description
When transferring bytes data into a TypeScript server authoritative match and relaying it back the bytes data gets corrupted on the process

## Steps to Reproduce
Send bytes data from Unity.
`nakamaManager.Socket.SendMatchStateAsync(match.Id, (long)code, bytes);`

Relay the message back to unity from the server.
`dispatcher.broadcastMessage(message.opCode, message.data, null, message.sender);`

Read bytes data received in Unity.

## Expected Result
The byte data received in Unity should be exactly the same as the sent data

## Actual Result
I sent this from unity: (Bytes to string with a space between each one)
12 0 0 0 8 0 16 0 14 0 4 0 8 0 0 0 253 96 66 65 142 250 66 191 0 0 20 0

I received this after relaying the message: (Bytes to string with a space between each one)
12 0 0 0 8 0 16 0 14 0 4 0 8 0 0 0 239 191 189 96 66 65 239 191 189 239 191 189 66 239 191 189 0 0 20 0

## Context
Unity + TypeScript server authoritative

## Your Environment
Nakama 3.4.0-r1
Unity Nakama 3.0.0 package

sesposito:
Do not convert data to a string before passing the messages to the runtime,
this can cause encoding issues in the goja vm if the content is binary data.

Introduce js runtime helper functions binaryToString and stringToBinary to
facilitate conversion of message data to string.

Resolves #659

### DIFF
--- a/server/runtime_javascript_nakama.go
+++ b/server/runtime_javascript_nakama.go
@@ -39,6 +39,7 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/dop251/goja"
 	"github.com/gofrs/uuid"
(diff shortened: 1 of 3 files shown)


server/runtime_javascript_nakama.go
```diff
@@ -250,6 +251,42 @@ func (n *runtimeJavascriptNakamaModule) mappings(r *goja.Runtime) map[string]fun
...
+	}
+}
+
+func (n *runtimeJavascriptNakamaModule) binaryToString(r *goja.Runtime) func(goja.FunctionCall) goja.Value {
+	return func(f goja.FunctionCall) goja.Value {
+		if goja.IsUndefined(f.Argument(0)) || goja.IsNull(f.Argument(0)) {
+			panic(r.NewTypeError("expects a Uint8Array object"))
+		}
+
+		data, ok := f.Argument(0).Export().([]byte)
```
redbaron, line 265: is it going to be ([]byte) and not https://pkg.go.dev/github.com/dop251/goja#ArrayBuffer ?

sesposito: It looks like it would be wrapped by an `ArrayBuffer` if a typed array object would be created in js, but because we're passing the function a `[]byte` coming from Go directly it is represented as such, without the wrapping.

redbaron: It is an `Array` with integeres elements then, not `ArrayBuffer`. idk where `MatchDataMessage` comes from, assuming it is from the user code sending `Uint8Array`,  and then seeing it in their match loop as JS `Array` will probably be a bit confusing.

It is possible to achieve symmetry by constructing `Uint8Array` from `[]byte` , like following:

```
dataUint8Array, err := vm.New(vm.Get("Uint8Array"), vm.ToValue(vm.NewArrayBuffer(msgData)))
if err != nil {
  panic(err)
}
msgMap["data"] = dataUint8Array
```

sesposito: Yep good point, I've made the change so that we pass a `Uint8Array` to the js runtime.

zyro (approved):

vlob1:
Hi! This code does not work:
`const buf = new Uint8Array([1,2,3]);
    dispatcher.broadcastMessage(1234, buf, null, null);`

error message is:
`TypeError: expects data to be an Uint8Array, a string or nil`

binaryToString throws same

PR #736, opened 2021-12-13
introduce batch processing to the Matchmaker

mschoch:
Previously, all updates to the Matchmaker index were done
immediately without any batching.  This causes excessive
overhead in the Bluge index, and significant gains are made by
introducing batch indexing.

It is believed that correctness is maintained with this change
but testing is on-going and additional tests may be proposed
in separate commits.


mschoch:
DO NOT MERGE (work in progress)

Yanovskiy:
@mschoch any update?

zyro:
This is being superseded by other ongoing work.

PR #738, opened 2021-12-14
Console account lookup by provider or custom ids

sesposito:
Update textbox hint and package ui


novabyte:
A couple of notes on this change:

- We should break the lookup into two queries only I think. After some thought I think the concurrent queries will just consume more of the connection pool than needed to perform this search. The first query should be adjusted to be an exact match on user ID or a wildcard search with `ILIKE` on username.
- The second query can be an exact match on the social provider IDs with OR statements. Executed sequentially after the first query is made and we aggregate the results into what's displayed in the UI.

redbaron:
> Executed sequentially after the first query is made

or `UNION` them in a  single query

zyro (approved):

zyro:
Merged in ae21db7f9bf0bf84a49d57f8b60f0dac402a64c8 and will be part of the next release.

PR #741, opened 2021-12-20
Fix missing reference in match registry.

mkieweg:
There is a bug present in the current version of `match_registry.go` preventing the creation of authoritative matches in certain cases. If the `params` contain the `server.MatchmakerEntry` implementation of the `runtime.MatchmakerEntry` interface the decoding in `CreateMatch` fails because the wrong type is registered (`server.MatchmakerEntry` instead of `*server.MatchmakerEntry`, see [this](https://stackoverflow.com/a/54767962) Stackoverflow answer for more details.)

This PR entails extended testing to show the bug as well as a fix in two different commits for easy reproduction.

### DIFF
--- a/server/match_registry.go
+++ b/server/match_registry.go
@@ -41,7 +41,7 @@ func init() {
 	gob.Register(map[string]interface{}(nil))
 	gob.Register([]interface{}(nil))
 	gob.Register([]runtime.MatchmakerEntry(nil))
-	gob.Register(MatchmakerEntry{})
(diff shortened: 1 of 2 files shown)


novabyte:
Thanks for the pull request @mkieweg. I've assigned an engineer to review it who will look at it in the new year. Have a great holiday season! 🎄

zyro (approved):

zyro:
@mkieweg Nice find and thanks for the fix! This is now merged and will be part of the next release. 👍

PR #756, opened 2022-01-20
Remove deprecated AWS Marketplace builder

michal-harish:


build/aws-marketplace/cloudbuild.yaml
```diff
@@ -19,6 +19,6 @@ steps:
       - '-ec'
       - |
         source ~/.heroic-ecr-secret && \
-        docker tag nakama-awsmp:${TAG#v} $_MAP_ECR_REPOSITORY:$TAG
+        docker tag nakama-awsmp:${TAG#v} $_MAP_ECR_REPOSITORY:${TAG#v}-m1
```
mofirouz, line 22: What's this change?

michal-harish: That's versioning of the images since we can't replace them in the mp ecr.

mofirouz: I think we leave the old images as they are - and we just publish Nakama images that don't exist yet.

mofirouz: Alternatively we use the `r` suffix as we've used everywhere else before that stands for `revision`. I don't know what `M` stands for here?

mofirouz: Ok that's fine, let's use `-m` suffix :)

PR #765, opened 2022-01-26
Make the JS global object immutable

sesposito:
Prevent global variables and registered functions/objects from being reset in the JavaScript vms.


server/runtime_javascript.go
```diff
@@ -1955,3 +1956,18 @@ func evalRuntimeModules(rp *RuntimeProviderJS, modCache *RuntimeJSModuleCache, m
 
 	return initializer.Callbacks, initializer.MatchCallbacks, nil
 }
+
+// Equivalent to calling freeze on the JavaScript global object making it immutable
+// https://github.com/dop251/goja/issues/362
+func freezeGlobalObject(r *goja.Runtime) {
+	r.RunString(`
+for (const k of Reflect.ownKeys(globalThis)) {
```
redbaron, line 1964: doesn't it freeze global variables instead of global object itself?

redbaron:
Having a test demonstrating that no new global variables can be created and ones which existed after module initialization are frozen would be nice.

Also does freezing disable writing to [Map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

redbaron:
This is a breaking change, probably worth having a very visible entry in the Changelog for it.

zyro (approved):

PR #771, opened 2022-02-01
Updates javascript runtime stream annotations

tomglenn:


seanpackham:
@tomglenn just checking, there are a few lines removed for each of those functions. Is that correct?

PR #772, opened 2022-02-01
Switch order of http_key and Authorization header auth methods

jasonwinterpixel:
We are working on integrating XSolla's payment system into our game.

XSolla's payment (and other systems) work by sending an API call to our server with a shared key and product information.

We'd like to handle that webhook in our nakama RPC system, but currently we cannot because XSolla is attaching a header to the request of the form: Authorization: Signature <a bunch of characters> 

Nakama is attempting to parse that Authorization header into a user's token and is not allowing the webhook through.

Prioritizing http_key seems to make sense to me and will allow us to integrate webhooks from other 3rd party services into our Nakama backend.

The diff looks more complicated than it is. It's literally just swapping the order of the if/else if

### DIFF
--- a/server/api_rpc.go
+++ b/server/api_rpc.go
@@ -51,7 +51,18 @@ func (s *ApiServer) RpcFuncHttp(w http.ResponseWriter, r *http.Request) {
 	var username string
 	var vars map[string]string
 	var expiry int64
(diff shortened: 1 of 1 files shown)


zyro (approved):
This looks like a fair change, thanks! 👍

Issue #537, opened 2021-01-26
Allow RPC registration without authentication


thegoldenmule:
#### Summary

We would like to be able to register RPCs that bypass authentication and are publicly accessible. Currently, all RPC registered through `RegisterRpc` require a user to be logged in or an HTTP call (for S2S calls). However, there are many use cases for un-authenticated clients to hit RPC endpoints.

Here are a few examples:

1. Nakama serves localization versioning information, which the client needs to be able to find the appropriate login button.
2. Users are broken into cohorts based on device id, before login occurs. This is fairly standard to test different first time user experience flows.
3. Before a user creates an account, they can see a live leaderboard to excite them about playing.
4. Releasing a game-specific public API for developers, hobbyists, and tinkerers (see [Fortnite API Example](https://fortnite-api.com/v1/map)).

### Workarounds

There are a few different workarounds we have considered:

- Making a new server that effectively proxies these requests. It has a public API, then sends requests along with the HTTP key to Nakama. This, unfortunately, is a time-consuming and resource-intensive alternative. We would need to deploy and scale this service ourselves rather than rely on Heroic Cloud's builtin clustering, which solves the same problem.

- Including the HTTP key on the client and making our own secret. This effectively compromises the HTTP key and opens every HTTP RPC endpoint for public access. We would then need to go through all of these and make sure the relevant check exists. The issue with this approach is not so much the work of adding checks, but making sure they stay in place. In many server frameworks (Rails, Express, ASP.NET, etc), this would be applied as policy middleware, and could be enforced across the board to avoid accidental security breaches. Nakama, to my knowledge, doesn't support this type of middleware strategy, so it would be up to us, every time we author an RPC function, to add that security check. Additional boilerplate for every RPC function is non-optimal in a number of ways.

#### Conclusion

These methods have significant drawbacks and it would be optimal to provide this functionality as part of Nakama. We are open to work with Heroic and contribute, but don't want to spend time on a PR if it will be rejected out of hand for some other reason.

novabyte:
@thegoldenmule Thanks for the detailed feature request. I think these requirements are very reasonable and will have a discussion with the team internally to determine whether there's tradeoffs I've missed. I'll leave a note here once that's happened for us to push it forwards 👍

thegoldenmule:
@novabyte Thanks for working with us. Again, we are happy to put some engineering resources on this so please let me know how we can help out. I understand that having us do the work, specifically for a _security_ related feature is dicey.

[...]

PR #775, opened 2022-02-03
add rest interface with default raw json content type

jordo:
This adds a by default unwrapped `/v2/rest` route, which routes to the same RPC endpoints.  This was needed by us for some integrations (Xsolla purchsing for one) that need to connect to endpoints with unwrapped application/json content-type requests, but are unable to add the `unwrap=X` URI parameter.


BergerBytes:
Just a bystander, but was there a reason this wasn't accepted? Is there anything that needs to be done first?

jordo:
Yes, this changeset is pretty minimal.  ping @novabyte

zyro:
See [this issue comment](https://github.com/heroiclabs/nakama/issues/537#issuecomment-2447469644) for updates. For the moment an alternative approach has been necessary to handle all requirements.

PR #783, opened 2022-02-08
Create ts client code generator from swagger.json file

ftkg:
Based on https://github.com/heroiclabs/nakama-js/tree/master/openapi-gen:

- adjusted to angular projects, mirroring the currrent clients
- added fixes and new schema elements

Differences to current client:
- already url encodes all non-path parameters
- accepts prefixes to remove as a command line flag, because swagger adds the name of imports/the file itself as prefixes to types and functions
- swagger does not bring types used on path+body POST requests, so these POST functions now accept a body argument as an anonymous object of the required fields
- map types are mapped as Map<>, not as type arrays
- swagger does not bring the enum type on arguments, so it needs to be passed as enum.ToString()

Resolves heroiclabs/mc2#476


ftkg:
> LGTM - does this change anything in Nakama Console?

There are no structural changes asides from a few missing types (other ones such as naming is being dealt by the generator), so in practice it should keep working the same as previously, as I've lightly tested it.

(what changed on console: updated `build.go`, `console.service.ts`, and adapted API calls on some client classes for missing types)

mofirouz:
I think let's generate the new stubs for the console and open a separate PR plz.

mofirouz:
I think there is a small bits that need changing in the generator - let's speak offline together. I also think we need to go through the entire Console and make sure all functions are working given the change here.

mofirouz (approved):

mofirouz:
Please let's roll this into Nakama OSS Console too!

PR #786, opened 2022-02-12
Seed global random using crypto/rand

shpikat:
Currently the global rand is seeded using `time.Now().UnixNano()`. That approach has its` drawbacks:

* The randomness is presented in the lower bits only, whereas the higher bits remain the same and very predictable.
* Some systems do not provide the granularity of single nanoseconds, sometimes it’s up to tens of milliseconds. That also increases predictability significantly.

Although for cryptographically strong random values the `crypto/rand` must be used, that’s not always the case where it’s needed. And some innocently looking predictable values may lead to serious exploits. Like jitter value for retries back-off.

Opening a PR as suggested in the discussion: https://forum.heroiclabs.com/t/better-seeding-of-global-rand/2354

The seeding procedure was pushed down the execution flow in order to use `startupLogger` to notify about the seed value fallback. That may influence any global rand usage at the early stages of the startup.

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,6 +16,8 @@ package main
 
 import (
 	"context"
+	cryptorand "crypto/rand"
+	"encoding/binary"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


zyro:
@shpikat Thanks for this - nothing before the new global `rand` seeding is expected to call on it for a random value, so it's safe to move this to after logger initialisation.

zyro (approved):

PR #787, opened 2022-02-14
Change httpKey to http_key in swagger file.

jasonwinterpixel:
Related: https://github.com/heroiclabs/nakama/pull/665
https://github.com/heroiclabs/nakama-godot/issues/89
https://github.com/heroiclabs/nakama-godot/pull/90

The swagger file has httpKey when it is actually http_key.

### DIFF
--- a/apigrpc/apigrpc.swagger.json
+++ b/apigrpc/apigrpc.swagger.json
@@ -2549,7 +2549,7 @@
             "type": "string"
           },
           {
-            "name": "httpKey",
+            "name": "http_key",
             "description": "The authentication key used when executed as a non-client HTTP request.",
             "in": "query",
             "required": false,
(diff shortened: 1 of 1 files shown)


dsnopek:
It looks like this is the commit where 'http_key' turned into 'httpKey' in the swagger file: https://github.com/heroiclabs/nakama/commit/0ff7d3581c1ed90c7372e9f399bbd27462d34821

This appears to have been done for several other query arguments as well, so it seems to be intentional?

mofirouz:
@dsnopek I think the bigger issue is that we don't have a wrapper API in the client library that wraps over the generated API. We've done this on all other SDKs (minus Defold) to prevent these issues.

Having said this, what happened in this commit is likely an accidental breaking change - but that's a separate discussion to whether we should keep it (because Swagger Gen has been changed, and we now have v3.10 is already released).

dsnopek:
@mofirouz Understood! As I noted [here](https://github.com/heroiclabs/nakama-godot/issues/89#issuecomment-1057341133), it looks like the code generator for the .NET library is converting the query params to snake case, which I think is what the code generator in the Godot library should do too. I'm planning to implement that shortly :-)

PR #800, opened 2022-03-03
Add condition healthcheck on Dockerfile

tunglt1810:
- Add version 3.8
- Add dependency condition with health check


zyro:
@tunglt1810 Could you please provide a bit more context on the rationale for this change?

tunglt1810:
Hi @zyro,

I'm using Macbook pro 2021 with M1 Pro chip.

When i'm using Docker desktop with current docker-compose config with cockroachDB, the game server usually stop at the first time i start the container. Because of the game server try to connect DB when it's not ready. And to start game server, i need to manually start the game server again.

So with this change, docker will wait the db start and ready with healthcheck and start game server

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3'
+version: '3.8'
```
sesposito, line 1: I suggest we keep this at `3` so we don't require the latest version as `condition` is supported anyway.

sesposito (changes requested):
Thank you @tunglt1810, this is a welcome addition, however I left a minor comment. Could you please also make the same changes to `docker-compose-postgres.yml` please?

PR #813, opened 2022-03-21
Improve StorageWrite to return acks in the same order as input objects

### LINKED ISSUE #754: Make StorageWrite return acks in the same order as input objects
[StorageWrite](https://github.com/heroiclabs/nakama/blob/cddb34e751288184ad5af9ed05e2dcdb58699367/server/runtime_go_nakama.go#L1862) doesn't guarantee that order of acks matches order of requested writes. It would make API easier to use correctly if order was preserved

ftkg:
Sorting of input was moved to an internal function, so that `MultiUpdate` sees the same behavior when writing to Storage.

Resolves #754

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -45,6 +45,7 @@ type storageCursor struct {
 type StorageOpWrites []*StorageOpWrite
 
 type StorageOpWrite struct {
+	Index   int
 	OwnerID string
 	Object  *api.WriteStorageObject
 }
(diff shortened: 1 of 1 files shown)


zyro (approved):

server/core_storage.go
```diff
@@ -458,9 +458,6 @@ WHERE
 }
 
 func StorageWriteObjects(ctx context.Context, logger *zap.Logger, db *sql.DB, authoritativeWrite bool, ops StorageOpWrites) (*api.StorageObjectAcks, codes.Code, error) {
-	// Ensure writes are processed in a consistent order.
-	sort.Sort(ops)
```
redbaron, line 462: @zyro, @ftkg , sorting prevents deadlocks. I suspect it also somehow helped CockroachDB in clustered mode.

zyro: Good call, fixed and updated comments to ensure we don't forget.

PR #819, opened 2022-03-25
Generate openapi spec and the nakama console client using the original proto names

ftkg:
This will use the original proto names for generating OpenAPI definitions (stop it from converting to camelCase), and accordingly adjusts the generator to stop converting case styles.

It should avoid issues where a different case style from the expected (generator always assumed snake_case) was being used in the source proto, thus breaking the generated client.


mofirouz:
@zyro let's have a quick chat about this offline.

PR #825, opened 2022-03-30
Vendor the dependencies

flaviofernandes004:


flaviofernandes004:
I was using an old golang version. closing this

PR #826, opened 2022-03-31
Add Console logger format option

ftkg:
Adding a human-readable format that allows plain-text messages to be logged.


novabyte:
@ftkg I think we can skip this addition and advise on how to use `jq` to pipe logs and transform them if developers want to view logs differently locally. All machine collection systems that do log aggregation benefit from the JSON format so this won't be used as part of production settings.

An example for anyone curious about what simple `jq` output looks like:

```shell
./nakama --database.address "$USER@127.0.0.1:5432?sslmode=disable" | jq -r '[.level, .ts, .msg?] | @tsv'
```

It will produce output which looks like:

```shell
info	2022-03-31T22:47:49.417+0100	Nakama starting
info	2022-03-31T22:47:49.417+0100	Node
info	2022-03-31T22:47:49.417+0100	Data directory
info	2022-03-31T22:47:49.417+0100	Database connections
debug	2022-03-31T22:47:49.418+0100	Complete database connection URL
info	2022-03-31T22:47:49.432+0100	Database information
info	2022-03-31T22:47:49.443+0100	Initializing leaderboard rank cache
debug	2022-03-31T22:47:49.443+0100	Caching leaderboard ranks
info	2022-03-31T22:47:49.445+0100	Initialising runtime
```

PR #828, opened 2022-04-05
Update angular generator

ftkg:


flaviofernandes004:
please include this [commit](https://github.com/heroiclabs/nakama/pull/819/commits/de66cfb4b0d2a22dc321a2064e36694a96723f44) to make the installation process more intuitive on other systems

console/ui/src/app/console.service.ts
```diff
@@ -1183,11 +1183,11 @@ export class ConsoleService {
   }
 
   /** List (and optionally filter) storage data. */
-  listStorage(auth_token: string, userId?: string, key?: string, collection?: string, cursor?: string): Observable<StorageList> {
+  listStorage(auth_token: string, user_id?: string, key?: string, collection?: string, cursor?: string): Observable<StorageList> {
 		const urlPath = `/v2/console/storage`;
     let params = new HttpParams();
-    if (userId) {
-      params = params.set('userId', userId);
+    if (user_id) {
```
mofirouz, line 1189: just to confirm; the changes done here (which are what we agreed on our call); do they affect the StorageList object, and therefore what the dashboard TypeScript code relies on? if not, then this PR LGTM.

mofirouz (approved):
One question, otherwise LGTM.

console/ui/src/app/console.service.ts
ftkg: Nope, these query params are passed as function args, so nothing changes for the caller:

`listStorage(auth_token: string, user_id?: string, key?: string, collection?: string, cursor?: string)`

PR #837, opened 2022-04-20
Adds a flag for disabling the console

orta:
Hi folks, one of my test server environment has quite a limit on the values of exposed ports (and a limit on them) and I wondered if being able to disable the console as a config option would be viable upstream? 

( This is also my first PR to a go project, so I'm not sure if I've missed something too! )

### DIFF
--- a/main.go
+++ b/main.go
@@ -164,7 +164,11 @@ func main() {
 	statusHandler := server.NewLocalStatusHandler(logger, sessionRegistry, matchRegistry, tracker, metrics, config.GetName())
 
 	apiServer := server.StartApiServer(logger, startupLogger, db, jsonpbMarshaler, jsonpbUnmarshaler, config, socialClient, leaderboardCache, leaderboardRankCache, sessionRegistry, sessionCache, statusRegistry, matchRegistry, matchmaker, tracker, router, metrics, pipeline, runtime)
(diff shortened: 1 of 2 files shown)


zyro:
@orta Thanks for the PR! I'm a bit wary of introducing this option to support what sounds like quite a niche environment. Can you go into a bit of detail on what this environment is, its characteristics, and why for example you can't just leave that port bound by Nakama but inaccessible?

orta:
I get ya. I'm self-hosting on render.com (which allows services to only have [one accessible port](https://feedback.render.com/features/p/expose-more-than-one-public-tcp-port-on-web-services)). Render seems to arbitrarily pick one of the open ports (maybe the first opened) and that is the only one available externally. Which means playing roulette each time the server restarts as to whether I'll be connecting to the main API or the console. 

I'll see if I can work around this via a custom dockerfile which doesn't expose the console's ports

zyro:
@orta Exposing one random port out of those bound by the application seems like a very odd solution, I would expect some kind of configuration to allow you to choose or at least for you to be given a specific port you can use. 🤔 

To add to this, Nakama actually opens up to 5 ports: console GRPC, console GRPC Gateway, client GRPC, client GRPC Gateway, and Prometheus metrics exporter endpoint.

In light of both the port situation in Nakama, and the extremely niche environment this is intended to address, I'll close this PR as an unsuitable general-purpose solution. That said I'm open to seeing if we can address this some other way, I'll give this some thought and feel free to ping me if you have other ideas! 👍

orta:
No worries! 

After a re-examination I found a way to tell render which is the port that matters, so for those instances I expose only the websocket

PR #859, opened 2022-05-19
Fixes typo for handler name error logging

durango:


zyro (approved):
@durango Nice find, thanks for the PR!

Issue #860, opened 2022-05-20
Fatal concurrency error while starting Nakama (with lua runtimes) under heavy load


brianwinterpixel:
## Description

While a Nakama instance was restarting under heavy load, we encountered this fatal error:

```go
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 65087 [running]:
runtime.throw({0x1a541bc, 0xaaaaaaaaaaaaaa})
        runtime/panic.go:1198 +0x71 fp=0xc04e09f1c8 sp=0xc04e09f198 pc=0x992e91
runtime.mapassign_faststr(0x188e960, 0xc0004d4d20, {0xc000a940a0, 0x1e})
        runtime/map_faststr.go:294 +0x38b fp=0xc04e09f230 sp=0xc04e09f1c8 pc=0x97040b
github.com/heroiclabs/nakama/v3/server.newRuntimeLuaVM.func1(0xc000a940a0, {0xc000a940a0, 0x16}, 0xc04e26edc0)
        github.com/heroiclabs/nakama/v3/server/runtime_lua.go:2077 +0xf2 fp=0xc04e09f260 sp=0xc04e09f230 pc=0x16b4c92
github.com/heroiclabs/nakama/v3/server.(*RuntimeLuaNakamaModule).registerRPC(0xc0001e1320, 0x30)
        github.com/heroiclabs/nakama/v3/server/runtime_lua_nakama.go:297 +0x84 fp=0xc04e09f2b0 sp=0xc04e09f260 pc=0x16d02e4
github.com/heroiclabs/nakama/v3/server.(*RuntimeLuaNakamaModule).registerRPC-fm(0xc004446240)
        github.com/heroiclabs/nakama/v3/server/runtime_lua_nakama.go:285 +0x2c fp=0xc04e09f2d0 sp=0xc04e09f2b0 pc=0x17571ec
github.com/heroiclabs/nakama/v3/internal/gopher-lua.callGFunction(0xc004400370, 0x0)
        github.com/heroiclabs/nakama/v3/internal/gopher-lua/vm.go:202 +0x37 fp=0xc04e09f338 sp=0xc04e09f2d0 pc=0x1416037
```

<details><summary>Full trace:</summary>

```go
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 65087 [running]:
runtime.throw({0x1a541bc, 0xaaaaaaaaaaaaaa})
[...]

PR #863, opened 2022-05-20
thread safety for global callback table in js and lua runtimes

jordo:
Potential fix for https://github.com/heroiclabs/nakama/issues/860


jordo:
* not tested *

PR #864, opened 2022-05-20
register global callback and get callback are now thread safe

jordo:
Potential fix for https://github.com/heroiclabs/nakama/issues/860

Not tested


zyro:
Superseded by 2f99d5a7e70a0d4e0eed6bfd11129b4999057b8d, and will be part of the next release. FWIW in this scenario `sync.Map` is the ideal candidate due to few writes, no expected overwrites, and almost every operation being a read.

PR #878, opened 2022-07-01
Limit unsuccessful login attempts on the console

ftkg:
- Adds a cache of failed login attempts, providing both account-based and IP-based lockouts after a number of failed login attempts have been received in a determined amount of time.

- The user is locked from making any further login attempts until it automatically unlocks after a period of time.

- A successful login clears the failed attempts for that account, but cached IP failed attempts only clear after the specific timespan.

### DIFF
--- a/server/console_authenticate.go
+++ b/server/console_authenticate.go
@@ -75,7 +75,7 @@ func (s *ConsoleServer) Authenticate(ctx context.Context, in *console.Authentica
 	ip, _ := extractClientAddressFromContext(s.logger, ctx)
 	lockout, _ := s.loginAttemptCache.IsLockedOut(in.Username, ip)
 	if lockout != unlocked {
-		return nil, status.Error(codes.Unavailable, "Try again later.")
+		return nil, status.Error(codes.ResourceExhausted, "Try again later.")
 	}
 
 	role := console.UserRole_USER_ROLE_UNKNOWN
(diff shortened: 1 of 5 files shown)


zyro (approved):

server/console_authenticate.go
```diff
@@ -81,10 +86,20 @@ func (s *ConsoleServer) Authenticate(ctx context.Context, in *console.Authentica
 			role = console.UserRole_USER_ROLE_ADMIN
 			uname = in.Username
 			id = uuid.Nil
+		} else {
+			if lockout, until := s.loginAttemptCache.Add(s.config.GetConsole().Username, ip); lockout != LockoutTypeNone {
```
redbaron, line 90: @zyro , @ftkg ,

there is a bug in `extractClientAddressFromContext` allowing  IP address to  be spoofed, making it easy to bypass this limiter.  Bug is that IP address extracted as a first element of x-forwarded-for header, not rightmost (N - 1), where N is number of known proxies, as a result IP address as nakama knows it is entirely controlled by user. 

Even without IP spoofing, with enough IP addresses available it looks like it is quite possible to exhaust server memory  by authenticating with sufficiently log usernames.  If we truncate username to something like 20 chars before storing in the cache it will be much harder to pull it off

zyro: Per [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax) the left-most address is supposed to indicate the client address, assuming trustworthy proxies across the chain. Either way we'll leave IP-based restrictions for a later stage.

Account lockouts are only tracked for valid accounts that actually exist, a bad actor cannot fill up the list with arbitrary usernames.

PR #886, opened 2022-07-18
Log returned error from `RowsAffected`

dslizardo:
We encountered the following log `Could not write storage object, rowsAffected error.`. 

```
{
    "level": "debug",
    "ts": "2022-07-15T10:40:40.122Z",
    "caller": "server/core_storage.go:603",
    "msg": "Could not write storage object, rowsAffected error.",
    "query": "UPDATE storage SET value = $4, version = $5, read = $6, write = $7, update_time = now() WHERE collection = $1 AND key = $2 AND user_id = $3::UUID AND version = $8"
}
```

It doesn't happen very often but we would like to see if there any returned error here.

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -607,7 +607,7 @@ func storageWriteObject(ctx context.Context, logger *zap.Logger, tx *sql.Tx, aut
 		}
 		return nil, err
 	}
-	if rowsAffected, _ := res.RowsAffected(); rowsAffected != 1 {
+	if rowsAffected, err := res.RowsAffected(); rowsAffected != 1 {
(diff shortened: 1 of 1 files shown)


server/core_storage.go
```diff
@@ -607,7 +607,7 @@ func storageWriteObject(ctx context.Context, logger *zap.Logger, tx *sql.Tx, aut
 		}
 		return nil, err
 	}
-	if rowsAffected, _ := res.RowsAffected(); rowsAffected != 1 {
+	if rowsAffected, err := res.RowsAffected(); rowsAffected != 1 {
```
dslizardo, line 610: I'm not really sure if `rowsAffected != 1` returns an error but I think it won't hurt to try logging it.

zyro (approved):

zyro:
Thanks @dslizardo! There should be no issue capturing that error.

PR #892, opened 2022-07-26
Push net.Conn Deadline forward on each read.

redbaron:
SetDeadline sets absolute time after which all IO operations start to fail.

Previously deadline was pushed forward upon successful websocket message
receipt, which works for small messages, but fail to behave adequately
if large message is transmitted over a slow connection.

This change makes deadline to be pushed forward on every successful read,
allowing slow clients to complete transmission as long as they keep sending
bytes.

### DIFF
--- a/server/session_ws.go
+++ b/server/session_ws.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


lugehorsam:
I'd suggest that if we like the approach, then before merging this I'd test it with our cpp client.

zyro:
After observing further usage of the socket protocol over various projects, we've ultimately decided against this approach. It can be revisited later if needed.

PR #922, opened 2022-09-29
Optional token parameter on unlink functions

### LINKED ISSUE #366: Unlink facbook id from user account without oauth token
The current unlink facebook api requires a oauth token that matches the current user. With the facebook unity sdk, it pulls the user from the installed facebook app. If a user changes accounts in facebook, there's a mismatch between oauth token and current nakama user, and unlinking fails. 

feature request to make facebook token optional, so we unlink based on facebook and nakama user id instead.

ftkg:
Unlink functions will still accept the token parameter, providing the option of a more restrictive "check-if-exists" unlink action.

Resolves #366

### DIFF
--- a/server/core_unlink.go
+++ b/server/core_unlink.go
@@ -173,23 +173,23 @@ AND ((apple_id IS NOT NULL
 }
 
 func UnlinkFacebook(ctx context.Context, logger *zap.Logger, db *sql.DB, socialClient *social.Client, appId string, id uuid.UUID, token string) error {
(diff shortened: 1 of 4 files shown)


zyro (approved):

omarWW112233:
Can anyone help me with that?

zyro:
@omarWW112233 Can you clarify exactly what you mean? This change was just merged, it's not in any current release yet.

PR #936, opened 2022-10-20
fixed sql query missing param GetPurchaseByTransactionID

MaxenceDemers:
PurchaseGetByTransactionId currently scans 9 arguments but is only populated with 8 fields in the sql query.
Added the missing param to match full request.


sesposito (approved):
LGTM, thank you for the contribution.

PR #943, opened 2022-11-07
Split matchmaker.Process into smaller functions for readability

redbaron:
matchmaker.Process was one big hard to follow function. This PR attempts to extract isolated pieces of logic into own function.

It is probably best to review on commit by commit basis, rather than in PR diff interface.

All commits except last two are either trivial changes or result of IDE's refactor tool, so very mechanical in essence and chance of introducing bugs is minimal. Only manual input required in these refactors was to bubble up `break`/`continue` control statements from within function via bool return param.

Last two commits are more substantial:

- https://github.com/heroiclabs/nakama/pull/943/commits/ccf77ab5650583b9e7228237ce1648fe1b82737d (and preceeding https://github.com/heroiclabs/nakama/pull/943/commits/8de9f0fe822c581d9e13eb801c724161c64a2db7): unifies conflict check between indexes. We have 3 conflicts checks: hit -> current index,  each combo entry -> hit and hit -> each combo entry. All three are now done with the same function accepting 2 indexes and a boolean flag to check whether reverse conflict check is required.

- https://github.com/heroiclabs/nakama/pull/943/commits/19b9ecf59c14bf9d36b8f86f4203058d6a7ff315 : atomic unit of matchmaking is matchmaking index, but MM logic used to operate on entries, which required converting to index in multiple places. This commits makes MM loop to operate on indexes, producing final entries only at the very end.

### DIFF
--- a/server/matchmaker.go
+++ b/server/matchmaker.go
@@ -300,7 +300,7 @@ func (m *LocalMatchmaker) Process() {
 	indexCount := len(m.indexes)
 
 	defer func() {
-		m.metrics.Matchmaker(float64(indexCount), float64(activeIndexCount), time.Now().Sub(startTime))
+		m.metrics.Matchmaker(float64(indexCount), float64(activeIndexCount), time.Since(startTime))
 	}()
 
 	// No active matchmaking tickets, the pool may be non-empty but there are no new tickets to check/query with.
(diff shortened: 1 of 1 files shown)


zyro:
Closing for now. This will be handled as part of a larger upcoming matchmaker change. 🙇

PR #966, opened 2022-12-26
implement lua hot reload

deflinhec:
This feature reload script content from filesystem and apply to all lua runtime. 

Depends on the complexity of the script content, some of the script might failed.

If the reload process failed, the latest script content will be discard.
 
Here is a preview of the console ui modification.
Screenshot 2022-12-25 at 16 07 57

Logging a positive result.
Screenshot 2022-12-25 at 16 08 21


deflinhec:
Script example
```lua
local nk = require 'nakama'
----------------------------------------------------------------//
local function api(context, payload)
    local message, code
    repeat
        if context.user_id then
            if context.user_id ~= "" then
                code = -1; break
            end
        end
        message = {text="CHANGE_ME"}
    until true
    code = code or 0
    return code, message
end
----------------------------------------------------------------//
nk.register_rpc(function(context, payload)
    local code, message, err
    xpcall(function()
        code, message = api(context, payload)
    end, function(_err) err = _err end)
    message = message or {message = err}
    message.code = code or -1
    return nk.json_encode(message)
end, "hotfix")
----------------------------------------------------------------//
```

```bash
curl --location --request POST 'http://127.0.0.1:7350/v2/rpc/hotfix?http_key=defaulthttpkey&unwrap' \
--header 'Content-Type: application/json' \
--data-raw '{}'
```

PR #983, opened 2023-02-08
Fix storage update query.

redbaron:
Currently there are 2 unique indexes on the table:
```
"storage_pkey" PRIMARY KEY, btree (collection, read, key, user_id)
"storage_collection_key_user_id_key" UNIQUE CONSTRAINT, btree (collection, key, user_id)
```

Only latter trully constraint uniqueness, because it is a subset of former. When using ON CONFLICT, most specific index of these two should be used, otherwise if `read` column is updated it might attempt to insert into table violating unique constraint.


redbaron:
Superseeded by https://github.com/heroiclabs/nakama/pull/1059

PR #984, opened 2023-02-09
Demonstrator how we can get away with just single unique index

redbaron:
This shows how instead of 4 indexes (396GB) (all of which are unique, but not all of them are marked as such in DB):

```
    "storage_pkey" PRIMARY KEY, btree (collection, read, key, user_id)  (78GB)
    "collection_read_user_id_key_idx" btree (collection, read, user_id, key)  (122GB)
    "collection_user_id_read_key_idx" btree (collection, user_id, read, key) (122GB)
    "storage_collection_key_user_id_key" UNIQUE CONSTRAINT, btree (collection, key, user_id)  (74GB)
```

we can have just one unique (primary key) index of 74GB:

```
"storage_collection_user_id_key_idx" btree (collection, user_id, key)   (~74GB)
```

I reviewed every query to check that index will be used (did run some, but not all, for rest  assumed what PG will do). All requsts have collection fixed and many have `user_id` fixed as well, leaving `key` to be actually queried. where `user_id` is not fixed, it is used in the pagination key, so it is still using index. 

Main changes in this PR are:
- `read` column is gone from all indexes. 
- Pagination cursor uses only ORDER BY columns. I verified with PG11 it continues to use index if leading index columns like `collection` are fixed by WHERE, so no need for them to be in the cursor
- StorageListObjectsAll returns objects ordered by `(collection,user_id,key)` instead of `(collection, read, key, user_id)`. I dont think we guarantee order of returned objects as long as we return all of them it is not a breaking change, but it allows us to removes indexes

Not mergeable AS-IS because migration need to be more sophisticated

### DIFF
--- a/migrate/sql/20230208224400-storage-optimize-idx.sql
+++ b/migrate/sql/20230208224400-storage-optimize-idx.sql
@@ -1,0 +1,6 @@
+ALTER TABLE storage
+  drop constraint storage_pkey, add primary key using index "storage_collection_key_user_id_key";
+
+DROP INDEX storage_pkey RESTRICT;
+
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS collection_user_id_key_idx ON storage (collection, user_id, key);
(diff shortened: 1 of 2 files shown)


zyro:
The primary key and index structure has changed as of 3.18.0 so this discussion is no longer quite relevant. We can revisit as needed later.

PR #993, opened 2023-03-08
Add satori integration

sesposito:
nakama-common related changes: https://github.com/heroiclabs/nakama-common/pull/116


internal/satori/satori.go
```diff
@@ -0,0 +1,437 @@
...
+		vErr.Errors |= jwt.ValidationErrorExpired
+		return vErr
+	}
+	return nil
+}
+
+func (s *SatoriClient) generateToken(id string) (string, error) {
+	timestamp := time.Now().UTC()
+	claims := sessionTokenClaims{
+		SessionID:  uuid.Must(uuid.NewV4()).String(),
```
zyro, line 89: Value here should be the empty string `""`, rather than a random value.

zyro (changes requested):

zyro (approved):

PR #997, opened 2023-03-17
Attempt to exchange Google's authorization code when given one instead of as a JWT

flaviofernandes004:


social/social.go
```diff
@@ -352,11 +416,32 @@ func (c *Client) CheckGoogleToken(ctx context.Context, idToken string) (*GoogleP
...
+		t, err := c.exchangeGoogleAuthCode(ctx, idToken)
+		if err != nil {
+			c.logger.Debug("Failed to exchange an authorization code for an access token.", zap.String("auth_token", idToken), zap.Error(err))
+			return nil, errors.New("google id token invalid")
+		}
+
+		c.logger.Debug("Exchanged an authorization code for an access token.", zap.Any("token", t), zap.Error(err))
+
+		profile := GooglePlayServiceProfile{}
+		if err := c.request(ctx, "google play services", "https://www.googleapis.com/games/v1/players/me?access_token="+url.QueryEscape(t.AccessToken), nil, &profile); err != nil {
```
flaviofernandes004, line 429: this requires the following scope: https://developers.google.com/games/services/web/api/players/get#auth

andreparodi-bandai:
I was able to authenticate correctly with version 0.11.1 of the google play unity plugin. So I'd say it's working. 

Couple of things regarding the config:
* It also requires client_secret
* I left the redirect_uris empty. Not sure how this is applicable to a game.

flaviofernandes004:
Awesome, thank you for taking a look. 

The content of the config `google_auth.crendentials_json` depends only on Google, I made no assumptions about it's schema. Configuration errors at that level will be logged and then it's up to the developer to adjust whatever is necessary in the Google's Console.

social/social.go
zyro: Let's make sure this gets documented. 👍

zyro (changes requested):

PR #1009, opened 2023-03-30
Lua hot reload

deflinhec:
Lua hot reload for runtime match and rpc.


dragonlobster:
@deflinhec would this hot reload work for typescript modules as well?

deflinhec:
> @deflinhec would this hot reload work for typescript modules as well?

Nope, this PR is only for lua.

zyro:
We are currently not comfortable integrating hot reloading into the server itself due to considerations around various existing running processes: match handlers, RPCs, before/after hooks etc. Our concern is it would be difficult to reason about which version of the code is running for any given execution context, relative to the value that hot reloading provides.

If hot reloading-like behaviour is needed, we recommend setting up a file watcher to restart the whole server process. Nakama's startup is very lightweight, so this generally does not take longer than a couple of seconds.

t1gu1:
@zyro 
I think in counter part, it would be a nice idea to add in the Doc how to setting up a file watcher to restart the whole server process.

For development it's pretty convenient.

PR #1011, opened 2023-04-04
Init and clone new configs

ftkg:


server/config.go
```diff
@@ -516,6 +519,8 @@ func (c *config) Clone() (Config, error) {
 		Leaderboard:      &configLeaderboard,
 		Matchmaker:       &configMatchmaker,
 		IAP:              &configIAP,
+		Satori:           &configSatori,
+		GoogleAuth:       &configGoogleAuth,
```
flaviofernandes004, line 523: No need to clone this value because it will be redacted here https://github.com/heroiclabs/nakama/blob/master/server/console_config.go#L55

zyro (approved):

Issue #412, opened 2020-03-07
400 bad request error When trying to call custom RPC


JunaidMkdn:
i am trying to access the rpc function i registered in nakama as "create_match_table"

the parameters that given to the functions are 
-> url = "http://127.0.0.1:7350/v2/rpc/create_match_table?http_key=defalutkey"
-> reqPayload is a struct of the form

`type CreateTableResponse struct {
	TableId              string  
}`

there is json labels on each fields in the struct like
`json=houseId,proto3"` for example

and reqPayload is a struct of the form 

`type CreateTableRequest struct {
	HouseId              string     
	TourId               string      
	GameLabel            string       
	TableConfig          *TableConfig 
}`

## Description
When trying to call custom RPC registered using sling client in go its failing and returning "400 with bad request "  error in the response from nakama

`func (c *Client) HttpPost(ctx context.Context, url string, reqPayload, resPayload interface{}) error {`

	var errorResponse interface{}

	resp, err := c.Post(url).BodyJSON(reqPayload).Receive(resPayload, &errorResponse)
	code := resp.StatusCode
	if err != nil || !(200 <= code && code <= 299) {
        //coming here the code is 400 
		return errors.WithStack(err)
	}

	if errorResponse != nil {
		return fmt.Errorf("nakama returned error: %#v", errorResponse)
	}

	return nil
}`

## Expected Result
Example: No error is thrown and results are returned by server.

## Actual Result

- Nakama:"Node","name":"nakama1","version":"2.7.0+d5e0b5bb","runtime":"go1.13","cpu":4,"proc":4}
- Database: "CockroachDB CCL v19.1.2 (x86_64-unknown-linux-gnu, built 2019/06/07 17:32:15, go1.11.6)
- Environment name and version: docker 
- Operating System and version: mac osx

mofirouz:
Hi 👋 I don’t think this is a bug within Nakama (we use GitHub Issues to track bugs or feature requests). Could you kindly ask your question over on the forums please:

https://forum.heroiclabs.com

Thanks 🙏

JunaidMkdn:
i am sorry , thank you for your consideration:)

PR #1025, opened 2023-05-18
fix lua modulo from  gopher-lua #412

deflinhec:
Reference https://github.com/yuin/gopher-lua/issues/412


sesposito:
Thank you for this contribution, we've taken the chance to upgrade the internal version of Gopher lua in https://github.com/heroiclabs/nakama/pull/1029 which superseeds this PR so we'll close it.

Issue #1026, opened 2023-05-19
Cannot use updated users' account password


ahay-agile6:
Before you submit an issue we recommend you drop into the [community forum](https://forum.heroiclabs.com) and ask any questions you have or mention any problems you've had getting started with the server.

Please provide as much information as you can with this issue report. If you believe it may be an issue with one of the client libraries please report it on their [own trackers](https://github.com/heroiclabs?utf8=%E2%9C%93&q=nakama%20AND%20sdk&type=&language=).

## Description
Hello,
I am creating users and authenticating their accounts using the AuthenticateEmail method of Nakama.
I can create a user using  the Nakama Unity SDK version 3.7.0
```
_session = await _client.AuthenticateEmailAsync(
                email,
                password,
                null,
                true,
                null,
                null,
                canceller.Token
            );
```
and log in to the application with the newly created user no problems.

However, when I go to the Nakama console, go to `http://localhost:7351/#/accounts` and choose the newly created user, go to the "Authenticate" tab, enter a new password for them - _**even if it's the same password I just created them with**_ - then attempt to log in to the application with the account that gets the password updated, I get the error "Invalid credentials".

This only happens when I _**update the user's password from the console**_.

## Steps to Reproduce

1. Register a new user using the `AuthenticateEmail` method
2. Log in to application - this should work even if you log out and back in
3. browse to `http://localhost:7351/#/accounts` (or equivalent) and find the newly created user. Click on their account.
4. Click on the "Authenticate" tab
5. In the "Password" field, type in a new password
6. Click "Save"
7. Go back to the application and attempt to log in to the account with the new password

## Expected Result
User can log in with the new password.

## Actual Result
User is presented with "Invalid credentials" error.

## Context

- [x] Unity
- [ ] Unreal
- [ ] Other

## Your Environment

- Nakama: 3.16.0
- Database: postgres
- Environment name and version:
- Operating System and version: docker

novabyte:
Thanks for the report @ahay-agile6. You might have found a bug with how we hash the stored password through the Console API. We'll take a look and drop a note here once we've identified it and resolved the problem.

PR #1027, opened 2023-05-19
Fix console account password change

sesposito:
Resolves https://github.com/heroiclabs/nakama/issues/1026.

### DIFF
--- a/server/console_account.go
+++ b/server/console_account.go
@@ -642,7 +642,7 @@ func (s *ConsoleServer) UpdateAccount(ctx context.Context, in *console.UpdateAcc
 		}
 	}
 
-	var newPassword string
+	var newPassword []byte
 	if v := in.Password; v != nil {
 		p := v.Value
 		if len(p) < 8 {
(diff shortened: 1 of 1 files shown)


ahay-agile6:
Hello! I am the author of #1026 .
I was excited to see this change, so I pulled this branch down locally, built, and ran it.
However, this still does not fix the issue 😭 
Is there a way to test that the password change updates as expected?

Thank you 🙂

sesposito:
Odd, this branch definitely seems to fix it for me, I've tested several times, I'll double check again.

sesposito:
I tested on postgresql myself too and followed the steps in the issue and the password appeared to be set correctly.

ahay-agile6:
Hi again!

I figured out the issue - it was on the Unity side: I was using the `Password` content type of the `InputField` and I was using the value of the display text, which were only asterisks, not of the **`InputField.text`** - so I was only sending "****" and not the real password 😊 

This fix is working both in the Nakama console API explorer and from the Unity client for me!

Thank you so much @sesposito 🙏I really appreciate your time and help.

sesposito:
No worries, I'm glad it's sorted! :)

zyro (approved):

PR #1039, opened 2023-06-21
Order leaderboard owner records

sesposito:


server/core_leaderboard.go
```diff
@@ -339,6 +340,26 @@ func LeaderboardRecordsList(ctx context.Context, logger *zap.Logger, db *sql.DB,
...
 	}
 
+	// Sort owner records according to leaderboard ordering
+	var sortFn func(i, j int) bool
+	if leaderboard.SortOrder == LeaderboardSortOrderAscending {
+		sortFn = func(i, j int) bool {
+			if ownerRecords[i].Score == ownerRecords[j].Score {
+				return ownerRecords[i].Subscore < ownerRecords[j].Subscore
+			}
+			return ownerRecords[i].Subscore < ownerRecords[j].Subscore
```
syhpoon, line 350: Hm, should it be something like `return ownerRecords[i].Score < ownerRecords[j].Score` instead? :thinking:

server/core_leaderboard.go
```diff
@@ -339,6 +340,26 @@ func LeaderboardRecordsList(ctx context.Context, logger *zap.Logger, db *sql.DB,
 		_ = rows.Close()
 	}
 
+	// Sort owner records according to leaderboard ordering
+	var sortFn func(i, j int) bool
+	if leaderboard.SortOrder == LeaderboardSortOrderAscending {
+		sortFn = func(i, j int) bool {
```
syhpoon, line 346: Cached records sorting fallbacks to comparing user ids in case scores and subscores are equal:
https://github.com/heroiclabs/nakama/blob/master/server/leaderboard_rank_cache.go#L50-L63

Just wondering if we want to replicate this logic here too?

server/core_leaderboard.go
sesposito: Good catch

server/core_leaderboard.go
sesposito: Yep that's a fair point, I'll do the same to keep consistency 👍

server/core_leaderboard.go
```diff
@@ -339,9 +340,37 @@ func LeaderboardRecordsList(ctx context.Context, logger *zap.Logger, db *sql.DB,
...
+			if ownerRecords[i].Score == ownerRecords[j].Score {
+				if ownerRecords[i].Subscore == ownerRecords[j].Subscore {
+					return ownerRecords[i].OwnerId < ownerRecords[i].OwnerId
```
syhpoon, line 349:
```suggestion
					return ownerRecords[i].OwnerId < ownerRecords[j].OwnerId
```

server/core_leaderboard.go
```diff
@@ -339,9 +340,37 @@ func LeaderboardRecordsList(ctx context.Context, logger *zap.Logger, db *sql.DB,
...
+		}
+	} else {
+		sortFn = func(i, j int) bool {
+			if ownerRecords[i].Score == ownerRecords[j].Score {
+				if ownerRecords[i].Subscore == ownerRecords[j].Subscore {
+					return ownerRecords[i].OwnerId > ownerRecords[i].OwnerId
+				}
+				return ownerRecords[i].Subscore > ownerRecords[i].Subscore
+			}
+			return ownerRecords[i].Score < ownerRecords[j].Score
```
syhpoon, line 363:
```suggestion
			return ownerRecords[i].Score > ownerRecords[j].Score
```

syhpoon (approved):

zyro (approved):

PR #1041, opened 2023-06-26
use last address in x-forwarded-for header as trusted edge proxy

jordo:
The first address in the `x-forwarded-for`  header is easily spoofable by any client...   We deploy nakama behind a reverse proxy which is going to append to this header as the last trusted edge proxy.  So I'm guessing anyone deploying nakama behind a reverse proxy would probably want to use the right-most field here as opposed to the left-most. 

See also:   https://adam-p.ca/blog/2022/03/x-forwarded-for/

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -518,7 +518,8 @@ func extractClientAddressFromContext(logger *zap.Logger, ctx context.Context) (s
 	md, _ := metadata.FromIncomingContext(ctx)
 	if ips := md.Get("x-forwarded-for"); len(ips) > 0 {
 		// Look for gRPC-Gateway / LB header.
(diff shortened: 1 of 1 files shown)


jordo:
Actually this ends up being incorrect for requests via the grpc interface.  since nakama spins up a gprcgateway internally that appends itself to the xff list, the client ip extraction on this interface should probably be the second last xff entry if more than 1.

jordo:
Removing as we just ended up solving this by rewriting the header, and using `X-Envoy-External-Address` in our case.   I still think probably by default nakama use the last address here instead of the first to avoid easy spoofing, but regardless it can be worked around by infrastructure.

PR #1043, opened 2023-06-30
fix broken unit tests

syhpoon:


internal/gopher-lua/_glua-tests/issues.lua
```diff
@@ -333,6 +333,7 @@ end
 test()
 
 --issue #331
+--[[
```
syhpoon, line 336: I had to comment this test because it was failing and after a spending some time trying to understand what it's doing I couldn't figure it out. Running some of the expressions from the test in a few different lua interpreters didn't yield the results the test code expects, so :shrug:

zyro (approved):

PR #1044, opened 2023-06-30
Use fixed number of query args in StorageReadObjects

redbaron:
When selecting variable number of objects it's better to keep number of SQL query arguments constant, otherwise query statistics explode, because from PostgreSQL perspective query with different number of arguments is a distinct query.

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -416,51 +416,65 @@ func storageListObjects(rows *sql.Rows, limit int) (*api.StorageObjectList, erro
 }
 
 func StorageReadObjects(ctx context.Context, logger *zap.Logger, db *sql.DB, caller uuid.UUID, objectIDs []*api.ReadStorageObjectId) (*api.StorageObjects, error) {
-	params := make([]interface{}, 0, len(objectIDs)*3)
+	collectionParam := make([]string, 0, len(objectIDs))
+	keyParam := make([]string, 0, len(objectIDs))
+	userIdParam := make([]uuid.UUID, 0, len(objectIDs))
 
-	whereClause := ""
-	for _, id := range objectIDs {
-		l := len(params)
-		if whereClause != "" {
-			whereClause += " OR "
-		}
+	// When selecting variable number of object we'd like to keep number of
+	// SQL query arguments constant, otherwise query statistics explode, because
+	// from PostgreSQL perspective query with different number of arguments is a distinct query
+	//
+	// To keep number of arguments static instead of building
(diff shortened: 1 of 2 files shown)


server/db.go
```diff
@@ -232,6 +233,27 @@ func ExecuteRetryable(fn func() error) error {
 	return nil
 }
 
+// ExecuteRetryablePgConn Retry functions that perform non-transactional database operations on PgConn
+func ExecuteRetryablePgConn(ctx context.Context, db *sql.DB, fn func(conn *pgx.Conn) error) error {
+	c, err := db.Conn(ctx)
+	if err != nil {
+		return err
+	}
```
redbaron, line 241:
```suggestion
	}
	defer c.Close()
```

server/core_storage.go
```diff
@@ -393,51 +394,67 @@ func storageListObjects(rows *sql.Rows, limit int) (*api.StorageObjectList, erro
...
+				logger.Error("Could not read storage objects. Unable to parse requested user_id", zap.Error(err))
+				return nil, err
 			}
-		} else if id.GetUserId() == "" {
-			whereClause += fmt.Sprintf(" (collection = $%v AND key = $%v AND user_id = $%v AND read = 2) ", l+1, l+2, l+3)
-			params = append(params, id.Collection, id.Key, uuid.Nil)
 		} else {
-			whereClause += fmt.Sprintf(" (collection = $%v AND key = $%v AND user_id = $%v AND (read = 2 OR (read = 1 AND user_id = $%v))) ", l+1, l+2, l+3, l+4)
-			params = append(params, id.Collection, id.Key, id.UserId, caller)
+			reqUid = uuid.Nil
```
zyro, line 444: Should not be needed, the empty value for a `uuid.UUID` is `uuid.Nil` already.

server/db.go
```diff
@@ -232,6 +233,28 @@ func ExecuteRetryable(fn func() error) error {
...
+	defer c.Close()
+	return c.Raw(func(dc any) error {
+		conn := dc.(*stdlib.Conn).Conn()
+	retry:
+		if err := fn(conn); err != nil {
+			var pgErr *pgconn.PgError
+			if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.SerializationFailure {
+				// A recognised error type that can be retried.
+				goto retry
```
zyro, line 250: Not a fan of the `goto`. Can we rewrite this more neatly, and also implement the max retries we've done elsewhere already?

zyro (changes requested):

zyro (approved):

PR #1048, opened 2023-07-06
Add Github action to build PRs

redbaron:


zyro (approved):

slonopotamus:
I wonder why you decided to ignore my #651 for _two years_ but instead merged this.

zyro:
@slonopotamus Apologies, we ran into a situation where having a basic build action wired would have helped greatly, so we whipped something together quickly without returning to the existing PRs. 🙇 

We're planning more actions especially around building releases after tags, but this is not ready yet.

Issue #1037, opened 2023-06-15
[console] In Nakama console storage request can't pass %2B character


kylee001:
in query string '+' is ' '
to display '+' we use '%2B' but nakama console can't pass query param with '%2B'

![image](https://github.com/heroiclabs/nakama/assets/124127312/83add0d9-8844-4b14-be95-6f0ea1516d7b)

![image](https://github.com/heroiclabs/nakama/assets/124127312/a8b679e6-1e86-44d2-8459-d23c0d691d48)

kylee001:
<img width="1138" alt="스크린샷 2023-07-07 오후 3 29 13" src="https://github.com/heroiclabs/nakama/assets/124127312/2863f48e-7d7c-4733-b041-e0788ee875b7">

in this function "+" is changed to " " maybe

we have to process several characters before this function
for example "+" --> "%2B"
`return this.httpClient.get<StorageList>(this.config.host + urlPath, { params: params, headers: this.getTokenAuthHeaders(auth_token) })`

I can't use key that include "+" character because this issue can I solve it?

PR #1052, opened 2023-07-10
Encode Angular HttpClient generated params in console.

ftkg:
See: https://github.com/angular/angular/issues/11058

Solves #1037


console/ui/src/app/console.service.ts
```diff
@@ -839,14 +839,14 @@ export class ConsoleService {
   /** Delete (non-recorded) all user accounts. */
   deleteAccounts(auth_token: string): Observable<any> {
 		const urlPath = `/v2/console/account`;
```
sesposito, line 841: There's a few places where the generated file has indendation misalignment, this is likely due to a spaces vs tabs issues in the template.

sesposito (changes requested):
Just a minor 💅  comment, otherwise LGTM

console/console.pb.go
sesposito, line 1: Is this change needed?

ftkg: Probably zero impact, but I'm using the defined `protoc-gen-go` version in `go.mod`

sesposito (approved):

PR #1056, opened 2023-07-13
Add storage engine index

sesposito:
Nakama common PR: https://github.com/heroiclabs/nakama-common/pull/124


server/storage_index.go
```diff
@@ -0,0 +1,504 @@
...
+
+type StorageIndex interface {
+	Write(ctx context.Context, objects StorageOpWrites)
+	Delete(ctx context.Context, objects StorageOpDeletes)
+	List(ctx context.Context, indexName, query string, limit int) (*api.StorageObjects, error)
+	Load(ctx context.Context) error
+	Start(runtime *Runtime)
+}
+
+// TODO: Return the modified values so that it can be propagated to other nodes in the cluster
```
zyro, line 41: Let's talk this through.

server/storage_index.go
```diff
@@ -0,0 +1,504 @@
...
+	db                    *sql.DB
+	indexByName           *MapOf[string, *storageIndex]
+	indicesByCollection   *MapOf[string, []*storageIndex] // TODO: Support multiple indices per collection
+	customFilterFunctions *MapOf[string, RuntimeStorageIndexFilterFunction]
+}
+
+func NewLocalStorageIndex(logger *zap.Logger, db *sql.DB, config []StorageIndexConfig) (StorageIndex, error) {
+	blugeCfg := BlugeInMemoryConfig()
+
+	idxByName := &MapOf[string, *storageIndex]{}
```
zyro, line 60: If these are only written once on startup, then let's use a standard `map`.

server/storage_index.go
```diff
@@ -0,0 +1,504 @@
...
+	indicesByCollection := &MapOf[string, []*storageIndex]{}
```
zyro, line 61: Same as above, standard `map` if possible.

server/storage_index.go
```diff
@@ -0,0 +1,504 @@
...
+		}
+
+		reader, err := idx.Index.Reader()
+		if err != nil {
+			si.logger.Error("Failed to get index storage reader", zap.Error(err))
+			return
+		}
+
+		count, _ := reader.Count() // cannot return err
+		idx.EntryCount.Store(count)
```
zyro, line 222: Do we need this stored count?

zyro (changes requested):

server/runtime_javascript_init.go
```diff
@@ -58,9 +58,9 @@ type jsMatchHandlers struct {
 }
 
 type RuntimeJavascriptCallbacks struct {
-	Rpc                            map[string]string
-	Before                         map[string]string
-	After                          map[string]string
+	Rpc                            *MapOf[string, string]
```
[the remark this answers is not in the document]
sesposito: From the docs:
```
// The MapOf type is optimized for two common use cases: (1) when the entry for a given
// key is only ever written once but read many times, as in caches that only grow,
// or (2) when multiple goroutines read, write, and overwrite entries for disjoint
// sets of keys. In these two cases, use of a MapOf may significantly reduce lock
// contention compared to a Go map paired with a separate Mutex or RWMutex.
```
I guess this would fall into (1), but I can change it anyway if you think it's best

server/storage_index.go
sesposito: // The MapOf type is optimized for two common use cases: (1) when the entry for a given
// key is only ever written once but read many times, as in caches that only grow,
// or (2) when multiple goroutines read, write, and overwrite entries for disjoint
// sets of keys. In these two cases, use of a MapOf may significantly reduce lock
// contention compared to a Go map paired with a separate Mutex or RWMutex.

I think this usage falls into (1), wdyt?

server/storage_index.go
sesposito: Good point, I'll remove it

server/runtime_lua_nakama.go
```diff
@@ -9730,6 +9771,56 @@ func (n *RuntimeLuaNakamaModule) channelIdBuild(l *lua.LState) int {
...
+// @summary List storage index entries
+// @param indexName(type=string) Name of the index to list entries from.
+// @param queryString(type=string) Query to filter index entries.
+// @param limit(type=int) Maximum number of results ro be returned.
+// @return objects(table) A list of storage objects.
+// @return error(error) An optional error value if an error occurred.
+func (n *RuntimeLuaNakamaModule) storageIndexList(l *lua.LState) int {
+	idxName := l.CheckString(1)
+	queryString := l.CheckString(2)
+	limit := l.CheckInt(3)
```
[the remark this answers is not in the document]
sesposito: I agree with making it optional, but since this listing is not paginated, do we think we should cap it at 100 results only?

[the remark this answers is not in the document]
zyro: Cap it, yes. We can always loosen the constraints later.

server/storage_index.go
zyro: We only ever write to the whole map once, not once per key. The map is filled at startup (no concurrent writes anyway) and then never updated.

zyro (approved):

PR #1058, opened 2023-07-14
Release swagger github pages on master

deflinhec:
Hi there, I came up with this github workflow and I think is very helpful, so I would like to contribute back to the repository. Basicly this workflow create a new branch with name default to `gh-pages`, which can be configure through Github Action, and setup a swagger-ui hosted by `github.io`.

Reference: https://docs.github.com/en/pages/quickstart

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,49 @@
+name: Deploy GitHub Pages
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


deflinhec:
By the way, if needed, I can also provide a PR of pushing image to Github Container Registry, which can then pull from `ghcr.io/heroiclabs/nakama`.

PR #1063, opened 2023-07-26
storage: switch to native UUID type in StorageWrite

redbaron:
StorageList already uses native UUID, this switches StorageWrite  from string to `UUID` type.


server/core_storage.go
```diff
@@ -47,7 +49,7 @@ type storageCursor struct {
 type StorageOpWrites []*StorageOpWrite
 
 type StorageOpWrite struct {
-	OwnerID string
+	OwnerID uuid.UUID
```
redbaron, line 52: the gist of this PR

ftkg (approved):

PR #985, opened 2023-02-09
Upgrade pgx to v5 and use trimmed sql-migrate

sesposito:

PR #1067, opened 2023-08-01
Upgrade pgx to v5 and use trimmed sql-migrate (Rebase:4bf4c19)

flaviofernandes004:
Use trimmed sql-migrate
Config and db creation refactor.

Original PR: https://github.com/heroiclabs/nakama/pull/985


sesposito (changes requested):
Left a minor comment but otherwise LGTM 👍

main.go
```diff
@@ -116,14 +137,21 @@ func main() {
...
-	// Global server context.
-	ctx, ctxCancelFn := context.WithCancel(context.Background())
-
-	db, dbVersion := server.DbConnect(ctx, startupLogger, config)
-	startupLogger.Info("Database information", zap.String("version", dbVersion))
+	db := server.DbConnect(ctx, startupLogger, config, false)
 
 	// Check migration status and fail fast if the schema has diverged.
-	migrate.StartupCheck(startupLogger, db)
+	conn, err := db.Conn(context.Background())
```
redbaron, line 143: looks like this leaks connection. needs `conn.Close()` on success and errors paths

server/console_channel.go
```diff
@@ -55,7 +55,7 @@ func (s *ConsoleServer) DeleteChannelMessages(ctx context.Context, in *console.D
 
 		var res sql.Result
 		var err error
-		if res, err = s.db.ExecContext(ctx, query, &pgtype.Timestamptz{Time: deleteBefore, Status: pgtype.Present}); err != nil {
+		if res, err = s.db.ExecContext(ctx, query, &pgtype.Timestamptz{Time: deleteBefore, Valid: true}); err != nil {
```
redbaron, line 58: AFAIK `time.Time` is natively supported for both args and scanning into

server/core_account.go
```diff
@@ -72,7 +73,7 @@ func GetAccount(ctx context.Context, logger *zap.Logger, db *sql.DB, statusRegis
 	var updateTime pgtype.Timestamptz
 	var verifyTime pgtype.Timestamptz
 	var disableTime pgtype.Timestamptz
-	var deviceIDs pgtype.VarcharArray
+	var deviceIDs pgtype.Array[string]
```
redbaron, line 76: there is `FlatArray[T]` which is just a newtype for `[]T` and doesn't support multidimentional arrays because of that, but otherwise should work just as well here

server/core_authenticate.go
```diff
@@ -62,7 +62,7 @@ func AuthenticateApple(ctx context.Context, logger *zap.Logger, db *sql.DB, clie
 	// Existing account found.
 	if found {
 		// Check if it's disabled.
-		if dbDisableTime.Status == pgtype.Present && dbDisableTime.Time.Unix() != 0 {
+		if dbDisableTime.Valid && dbDisableTime.Time.Unix() != 0 {
```
redbaron, line 65: `disable_time` column is `NOT NULL` in schema, so all these checks can be replaced with simple `time.Time` argument  and emptines checked with `.IsZero()`

sesposito: `.IsZero()` and `.Unix() == 0` are not equivalent, I'd rather not touch this code as it's working as intended

sesposito:
Superseded by: https://github.com/heroiclabs/nakama/pull/1218

PR #1069, opened 2023-08-03
Upgrade docker compose crdb to v23.1

sesposito:
The reworked queries introduced in https://github.com/heroiclabs/nakama/pull/1044 are incompatible with some older versions of crdb due to this bug: https://github.com/cockroachdb/cockroach/issues/60907 that has been fixed. 

Upgrading crdb resolves this.

Alternatively, we'd have to merge https://github.com/heroiclabs/nakama/pull/1066 to keep backwards compatibility.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   cockroachdb:
-    image: cockroachdb/cockroach:latest-v20.2
+    image: cockroachdb/cockroach:latest-v23.1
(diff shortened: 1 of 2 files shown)


novabyte:
@sesposito My vote is that we lift the minimum required CRDB version rather than provide a workaround for an older CRDB release. It's a good time to lift the minimum version required as we move to PGX v5 as well.

redbaron:
I vote for this over #1066

zyro:
I agree, this is the correct solution. Let's merge this and close the others.

zyro (approved):

PR #1077, opened 2023-08-31
Fix before request hook logger flag

maciejmrozinski:


sesposito (approved):
Thank you for the contribution @maciejmrozinski

PR #1084, opened 2023-09-11
use Last() implementation when calculating tournament deadlines

lugehorsam:
Ports https://github.com/gorhill/cronexpr/pull/28 to our internal version of `cronexpr`. This gives us the `Last()` method which allows us to infer the start time directly from the cronexpr rather than estimating it by invoking `Next()` twice.

- Fixes calculations of leaderboard and tournament times for rare cron expressions that don't execute at a fixed interval.
- Fixes calculations of leaderboard and tournament times for future tournaments.


server/core_tournament.go
```diff
@@ -737,22 +737,17 @@ func calculateTournamentDeadlines(startTime, endTime, duration int64, resetSched
...
-					startActiveUnix = s
-				}
-				break
-			}
+		var startActiveUnix int64
+		// check if we are landing squarely on the reset schedule
+		if resetSchedule.Next(t.Add(-1*time.Second)) == t {
+			startActiveUnix = tUnix
+		} else {
+			// otherwise assume the supplied time is ahead of the reset schedule.
```
lugehorsam, line 745: I believe this assumption was baked into the for loop of the previous code as well.

lugehorsam:
I added a test for an additional case I thought of but this is still good to review.

server/core_tournament.go
sesposito: Is this covering the case where `startActiveTime` is in the future?

server/core_tournament_test.go
```diff
@@ -0,0 +1,92 @@
...
+	startActiveUnix, endActiveUnix, _ := calculateTournamentDeadlines(startTime, 0, duration, sched, nowUnix)
+	nextResetUnix := sched.Next(nowUnix).Unix()
+
+	// 14 August 2023, 9:00:00
+	require.Equal(t, int64(1692003600), startActiveUnix, "Start active times should be equal.")
+	// 21 August 2023, 9:00:00
+	require.Equal(t, int64(1692608400), endActiveUnix, "End active times should be equal.")
+	// 14 September 2023, 9:00:00
+	require.Equal(t, int64(1694682000), nextResetUnix, "Next reset times should be equal.")
+}
```
sesposito, line 92: I'd suggest adding a test case for when startTime > now

server/core_tournament.go
lugehorsam: Seeing the leaderboard code, it looks like `0` should be returned if `startActiveTime` is in the future.

lugehorsam:
Ready for re-review.

syhpoon (approved):
:+1:

lugehorsam:
Will still need a 👍 from @sesposito  if it deserves it to unblock the merge

sesposito (approved):

PR #1085, opened 2023-09-12
Return error on empty Google receipt PackageName

sesposito:


iap/iap.go
```diff
@@ -283,6 +283,10 @@ func decodeReceiptGoogle(receipt string) (*ReceiptGoogle, error) {
 	if err := json.Unmarshal([]byte(unwrapped), &gr); err != nil {
 		return nil, errors.New("receipt is malformed")
 	}
+	if gr.PackageName == "" {
+		return nil, errors.New("receipt is malformed")
```
syhpoon, line 287: would it make the error more useful if we actually returned the reason (like in `package name is empty`) here?

syhpoon (approved):

iap/iap.go
sesposito: being purposefully generic as this field should always be present if the receipt is in the expected format, otherwise valid json is unmarshalled without error but the struct will have none of the expected fields set.

PR #1107, opened 2023-10-18
feat: add an option to disable the session_cache

delasource:
Adds the configurable option "enable_tokenwhitelist" (default false) that when enabled, requires tokens to be issued by the exact nakama instance. This is the current Nakama behaviour and some users (including me) complained about it, since a Server Reboot resulting in all clients had to manually reauthorize (i.e. login via the external auth provider).

Setting the config value "enable_tokenwhitelist" to false will skip all the checks to the so called "sessionCache" meaning tokens will be accepted by the server as long as they are valid (reboot safe).


UmarBhatPlaystrom:
🔥🔥🔥

zyro:
Should now be resolved in [v3.18.0](https://github.com/heroiclabs/nakama/releases/tag/v3.18.0). Rather than a configurable option, we've changed the nature of the cache itself to rely more heavily on the semantics of JWTs.

PR #1108, opened 2023-10-18
Proper check for registering Go Matchmaker Override

maciejmrozinski:
Go runtime is not registering Matchmaker Override functionality because of incorrect check


sesposito (approved):

sesposito:
Nice one @maciejmrozinski 👍

PR #1114, opened 2023-10-23
Move queries with variable number of args to a fixed number of args syntax

ftkg:
Queries with a variable number of arguments are treated as distinct queries from PostgreSQL perspective, and their statistics are tracked separately. This PR aims to improve query metrics' cardinality and, consequently, improve query stats monitoring. This is achieved by using a fixed syntax:

- `WHERE x in ($1,$2,$3, ...)` is replaced with `WHERE x = ANY($1::type_of_x[])`, with all values passed as a single array argument.
- `INSERT INTO t (a,b,c) VALUES ($1,$2,$3), ($4,$5,$6), ...` is replaced with `pgx.Batch`, with a single row insert per query.

### DIFF
--- a/server/console_channel.go
+++ b/server/console_channel.go
@@ -5,8 +5,6 @@ import (
 	"database/sql"
 	"net/url"
 	"sort"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/gofrs/uuid/v5"
@@ -66,20 +64,16 @@ func (s *ConsoleServer) DeleteChannelMessages(ctx context.Context, in *console.D
 		s.logger.Info("Messages deleted.", zap.Int64("affected", affected), zap.String("timestamp", deleteBefore.String()))
 	}
 	if len(in.Ids) > 0 {
-		params := make([]interface{}, 0, len(in.Ids))
-		statements := make([]string, len(in.Ids))
-		for i, id := range in.Ids {
-			idStr, err := uuid.FromString(id)
+		for _, id := range in.Ids {
+			_, err := uuid.FromString(id)
 			if err != nil {
 				return nil, status.Error(codes.InvalidArgument, "Requires a valid message ID.")
 			}
-			params = append(params, idStr)
-			statements[i] = "$" + strconv.Itoa(i+1)
 		}
-		query := "DELETE FROM message WHERE id IN (" + strings.Join(statements, ",") + ")"
+		query := "DELETE FROM message WHERE id = ANY($1)"
 		var res sql.Result
 		var err error
-		if res, err = s.db.ExecContext(ctx, query, params...); err != nil {
+		if res, err = s.db.ExecContext(ctx, query, in.Ids); err != nil {
 			s.logger.Error("Could not delete messages.", zap.Error(err))
 			return nil, status.Error(codes.Internal, "An error occurred while trying to delete messages.")
 		}
--- a/server/core_wallet.go
+++ b/server/core_wallet.go
@@ -23,8 +23,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
-	"strconv"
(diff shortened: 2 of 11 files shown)


server/console_channel.go
```diff
@@ -66,20 +64,16 @@ func (s *ConsoleServer) DeleteChannelMessages(ctx context.Context, in *console.D
...
+		for _, id := range in.Ids {
+			_, err := uuid.FromString(id)
 			if err != nil {
 				return nil, status.Error(codes.InvalidArgument, "Requires a valid message ID.")
 			}
-			params = append(params, idStr)
-			statements[i] = "$" + strconv.Itoa(i+1)
 		}
-		query := "DELETE FROM message WHERE id IN (" + strings.Join(statements, ",") + ")"
+		query := "DELETE FROM message WHERE id = ANY($1)"
```
redbaron, line 73: does it work with generic `database/sql` interface without converting to `pgx`?

server/core_wallet.go
```diff
@@ -216,8 +208,7 @@ func updateWallets(ctx context.Context, logger *zap.Logger, tx pgx.Tx, updates [
 				return nil, err
 			}
 
-			params = append(params, uuid.Must(uuid.NewV4()), userID, changesetData, update.Metadata)
-			statements = append(statements, fmt.Sprintf("($%v::UUID, $%v, $%v, $%v)", strconv.Itoa(len(params)-3), strconv.Itoa(len(params)-2), strconv.Itoa(len(params)-1), strconv.Itoa(len(params))))
+			batchLedger.Queue(queryLedger, uuid.Must(uuid.NewV4()), userID, changesetData, update.Metadata)
```
redbaron, line 211: `uuid.Must` seems unnecessary becasue `uuid.NewV4()` doesn't fail

server/console_channel.go
ftkg: Yes, this seemingly works fine

server/core_wallet.go
ftkg: `uuid.NewV4()` actually returns `(UUID, error)`

ftkg:
Replaced `pgx.Batch` with `INSERT INTO .. SELECT unnest()` to avoid having to sort input data (due to deadlock potential). The single statement may also be more efficient.

redbaron:
> Replaced `pgx.Batch` with `INSERT INTO .. SELECT unnest()` to avoid having to sort input data (due to deadlock potential). The single statement may also be more efficient.

Queries in Batch are executed in the same order they were added to Batch, which is in turn the same as in unnest, so deadlock potential should be the same.

As for efficiency, pgx uses prepared statement underndeath, so any request after first will be sending just args for the query, not the query itself. `pgx` will spend time hashinb query string (to match  prepared statement), but I think simplicity of Batch makes it more appealing rather than fairly unconventional `unnest` .

If string hashing becomes a problem, then explicit prepared statement can be used with Batch.

ftkg:
> > Replaced `pgx.Batch` with `INSERT INTO .. SELECT unnest()` to avoid having to sort input data (due to deadlock potential). The single statement may also be more efficient.
> 
> Queries in Batch are executed in the same order they were added to Batch, which is in turn the same as in unnest, so deadlock potential should be the same.
> 
> As for efficiency, pgx uses prepared statement underndeath, so any request after first will be sending just args for the query, not the query itself. `pgx` will spend time hashinb query string (to match prepared statement), but I think simplicity of Batch makes it more appealing rather than fairly unconventional `unnest` .
> 
> If string hashing becomes a problem, then explicit prepared statement can be used with Batch.

The only other difference I see is tracking stats for every single query (Batch) vs the single bulk query (unnest) but I don't really prefer one or the other, wdyt?

zyro (approved):

PR #1118, opened 2023-10-25
Ignore conflicting storage inserts with version="*"

ftkg:
Upsert with `DO NOTHING` will ignore insert on conflicting key, instead of failing. The intended upside is that transactions with conflicting inserts will not have to be retried anymore, saving unnecessary database processing. 

Results do not change, as the new query will still return no rows on conflict, hence the batch insert will be aborted with `ErrStorageRejectedVersion`.

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -699,6 +699,7 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 		query = `
 		INSERT INTO storage (collection, key, user_id, value, version, read, write, create_time, update_time)
 		VALUES ($1, $2, $3, $4, $5, $6, $7, now(), now())
+		ON CONFLICT (collection, key, user_id) DO NOTHING 
 		RETURNING read, write, version, create_time, update_time`
 
 		// Outcomes:
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -699,11 +699,11 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 		query = `
 		INSERT INTO storage (collection, key, user_id, value, version, read, write, create_time, update_time)
 		VALUES ($1, $2, $3, $4, $5, $6, $7, now(), now())
-		ON CONFLICT (collection, key, user_id) DO NOTHING 
+		ON CONFLICT (collection, key, user_id) DO NOTHING
 		RETURNING read, write, version, create_time, update_time`
 
 		// Outcomes:
-		// - NoRows - insert failed due to constraint violation (concurrent insert)
+		// - NoRows - insert was ignored due to constraint violation (concurrent insert or non-OCC existing row)
 	}
 
 	batch.Queue(query, params...)


server/core_storage.go
```diff
@@ -699,10 +699,11 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 		query = `
 		INSERT INTO storage (collection, key, user_id, value, version, read, write, create_time, update_time)
 		VALUES ($1, $2, $3, $4, $5, $6, $7, now(), now())
+		ON CONFLICT (collection, key, user_id) DO NOTHING
```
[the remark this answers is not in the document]
ftkg: Added new lines to return the row if the data being inserted is the same.

ftkg:
New query should return success for concurrent writes of identical objects. Had to use `ON CONFLICT UPDATE` to return a row for this scenario, that's an extra step but it should be better than retrying (additionally, the previous `WHERE NOT EXISTS` will avoid unnecessary updates for non-concurrent duplicate writes and keep the semantics of rejecting them).

server/core_storage.go
```diff
@@ -698,11 +698,15 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 		// Existing permission checks are not applicable for new storage objects.
 		query = `
 		INSERT INTO storage (collection, key, user_id, value, version, read, write, create_time, update_time)
-		VALUES ($1, $2, $3, $4, $5, $6, $7, now(), now())
+		SELECT $1, $2, $3, $4, $5, $6, $7, now(), now()
+			WHERE NOT EXISTS(SELECT 1 FROM storage WHERE collection = $1 AND key = $2 AND user_id = $3) -- avoids ON CONFLICT hitting pre-existing object, and immediately cancels insert
+		ON CONFLICT (collection, key, user_id) DO UPDATE
```
redbaron, line 703: `ON CONFLICT ` misses `WHERE` to check that conflictings row are in fact the same, otherwise this query just behaves as "last  concurrent write wins"

redbaron:
`UNION ALL LIMIT 1 ` variant doesn't unnecessary overwrite row just to return it and can be made to work by checking returned row on the Go side: If returned row is the same which is written, then write succeeds.

server/core_storage.go
ftkg: `WHERE` clause goes after `ON CONFLICT DO UPDATE SET`

redbaron (approved):

redbaron:
This is a behaviour change which needs to be prominently mentioned in release notes and docs. Previously it was possible to use this operation to create distributed lock/ leader election , now it is still possible but written value must be unique for each node

zyro:
After internal discussion, we determined the current behaviour is actually desirable: `"*"` must fail on any existing value even if fully identical. This allows Nakama to support use cases such as using a storage object as a lock or signal if needed.

We may however introduce a new OCC operator at some point that allows the write to succeed if the object already existed but the value was identical.

Issue #1124, opened 2023-10-31
TournamentRecordWrite adds to leaderboard cache even if there isnt't a new score


robtraynor:
## Description

I was getting an error where an incorrect Rank was being returned by TournamentRecordsList, and I think I've tracked it do to  
TournamentRecordWrite, updating the rank cache even if the score isn't used.

In the case where the score isnt updated and the leaderboard is doesnt require a join  ` rankCache.Insert`  is always called:
```
	record.Rank = rankCache.Insert(leaderboard.Id, leaderboard.SortOrder, record.Score, record.Subscore, oldScore, oldSubscore, expiryUnix, ownerId)

```
in core_tournament.go.

In this case oldScore and oldSubscore are zero, since they are set from the result of the database update which is empty in the case where there is no update. Meaning that when the rankCache looks to delete the old entry there are no matches and instead it is added. 

I can create pull request to fix if you want by adding logic similar to core_leaderboard.go and calling `rankCache.Get`, if there hasn't been an update, but im not sure if this logic would need to be mirrored in the requiredJoin branch for cosistency.

## Steps to Reproduce

1. Set up a tournament without required join and operator set to best
2. Use any client to call TournamentRecordWrite
3. Call TournamentRecordWrite again with a lower score

## Expected Result
The rankCache for the leaderboard should contain one entry

## Actual Result
The rankCache for the leaderboard contains two entries

zyro:
@robtraynor Nice find, this makes sense. The rank cache entries are definitely intended to be unique so if there's a situation that adds an entry without removing the previous one, that's not intended.

A PR would be awesome! 🙏

zyro:
Off the top of my head I think `max_num_scores` is meant to be the submitted count full stop, improved or not. I need to look into it to be sure. 👀 

That sounds like a separate issue though, right? If memory serves the number of scores doesn't affect rank and/or rank caching.

robtraynor:
I think if the score is not an improvement, the values passed to the rank cache for the old values will always be 0, since the insertion to the cache tries, therefore it cant find it to remove it from the cache and instead adds a new one. and since the insertion to the cache tries to since they are set from the result of query which fails due to the WHERE condtion which only updates the value if it's better. So I think while they are two seperate issues but caused by the same issue.

robtraynor:
@zyro Hey just wondering if there is an update?

flaviofernandes004:
Sorry, I closed this by mistake.

PR #1125, opened 2023-11-01
WriteTournamentScore now updates num_scores even if submitted score is not an improvement

robtraynor:
Previously if a score was not an improvement the WHERE clause on the update query would prevent the num_scores from being updated, this would also cause an issue with the rank cache not correctly deleting the previous score, since the values of the old scores are only retrieved if the query succeds, meaning values would get added multiple times.

This should fix https://github.com/heroiclabs/nakama/issues/1124


flaviofernandes004 (approved):

zyro:
Thanks @robtraynor! 🙇

PR #1128, opened 2023-11-08
Add option to disable relayed matches after matchmaker error, fixes for tests

maciejmrozinski:
I thought that feature to disallow automatic relayed match token generation after autoritative matchmaker error will be usefull.

This option set to false (with blocking before hook implementation on MatchCreate) will totally disable relayed matches, which can be desirable by some use cases.

I also fixed all errors in tests to get them all passed.


novabyte:
@maciejmrozinski Thanks for the contribution. Can you provide more detail on this point "which can be desirable by some use cases" - what are the desirable use cases to disable the default behaviour for relayed multiplayer matches?

maciejmrozinski:
Main point is to totally disable relayed match creation. Right now we can block it from client side by implementing guard function on before create match hook. But even then when we use matchmaker to create authoritative matches and error occurs, Nakama generates match token without control and sends it to clients. Probably we can block joining non-authoritative matches on before match join hook, but disabling token generation seems like more optimal solution.

zyro:
@maciejmrozinski I think it's important to note relayed matches aren't really "created" in the sense this word might imply. The matchmaker only generates match join tokens.

If you prevent (with a before hook) the socket operations that create matches or join matches with a token (not a match ID, that's only for authoritative matches) then no resources will be consumed on the server at all.

In a practical sense is what you're looking for more to prevent matchmaking succeeding/completing if creating authoritative matches fails? With or without reporting the issue to the client?

maciejmrozinski:
You're right that from other point of view describing this solution as a way to prevent matchmaking from completing and using ticket when creating authoritative match fails is also correct. Probably in base version, as it's implemented in PR now, not consuming a ticket so it can be picked in another matchmaker run is enough. In future, comunicating this event to the client maybe can be usefull.

PR #1132, opened 2023-11-09
Improve query planning in StorageReadObjects

ftkg:
This aims to optimize queries for cases where all requested storage objects are for the same user/key/collection:

1. For any requested column, if all requested values are the same then it becomes a simple `WHERE $column = $value` condition;
2. If among all requested columns only one has multiple distinct values requested, it becomes a `WHERE $column = ANY($values[])` condition;
3. Otherwise, if we have two or three columns with multiple distinct values requested, these columns become part of a `JOIN ROWS FROM ()` virtual table clause.

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -423,6 +423,10 @@ type storageQueryArg struct {
 }
 
 func StorageReadObjects(ctx context.Context, logger *zap.Logger, db *sql.DB, caller uuid.UUID, objectIDs []*api.ReadStorageObjectId) (*api.StorageObjects, error) {
+	if objectIDs == nil || len(objectIDs) == 0 {
+		return &api.StorageObjects{}, nil
+	}
+
 	collectionParams := make([]string, 0, len(objectIDs))
 	keyParams := make([]string, 0, len(objectIDs))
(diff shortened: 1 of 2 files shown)


server/core_storage.go
```diff
@@ -415,45 +416,45 @@ func storageListObjects(rows *sql.Rows, limit int) (*api.StorageObjectList, erro
...
+	isUserIdSetUnique := true
+
+	distinctArgs := make([]storageQueryArg, 0, 3)
+	uniqueArgs := make([]storageQueryArg, 0, 3)
+
 	for _, id := range objectIDs {
-		collectionParam = append(collectionParam, id.Collection)
-		keyParam = append(keyParam, id.Key)
+		collectionParams = append(collectionParams, id.Collection)
+		if isCollectionSetUnique && len(collectionParams) > 1 {
```
redbaron, line 443: len check is not necessary, `!= [0]` check on first value appended will fail

redbaron (approved):

sesposito (commented):
The changes look good but because this a critical code path I'd suggest we add some tests to cover the different built queries.

sesposito (approved):

PR #1135, opened 2023-11-10
Dockerfile improvements.

sesposito:
Upgrade containers to Debian 12.
Remove unneeded dependencies from release container. 
Add healthcheck through Nakama binary.

### DIFF
--- a/build/Dockerfile.local
+++ b/build/Dockerfile.local
@@ -12,13 +12,9 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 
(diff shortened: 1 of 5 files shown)


build/Dockerfile.local
```diff
@@ -0,0 +1,57 @@
+## Copyright 2018 The Nakama Authors
```
zyro, line 1: Do we need this file?

zyro (changes requested):

build/Dockerfile.local
sesposito: This is sometimes useful for development to build a Nakama container based on the local changes instead of pulling from the repository, but I can remove it if deemed unnecessary

zyro (approved):

PR #1146, opened 2023-11-22
Add safeguard around js caching api

sesposito:
Prevent js objects and other types to leak vm context through the localcache by disallowing such types to be cached and restricting to primitive data types only on put.

### DIFF
--- a/server/runtime_javascript_nakama.go
+++ b/server/runtime_javascript_nakama.go
@@ -8004,7 +8004,15 @@ func (n *runtimeJavascriptNakamaModule) localcachePut(r *goja.Runtime) func(goja
 			panic(r.NewTypeError("ttl must be 0 or more"))
 		}
 
-		n.localCache.Put(key, value.Export(), ttl)
+		v := value.Export()
+
+		switch v.(type) {
+		case string, int64, float64, bool:
+		default:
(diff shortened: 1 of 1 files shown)


server/runtime_javascript_nakama.go
```diff
@@ -8004,7 +8004,15 @@ func (n *runtimeJavascriptNakamaModule) localcachePut(r *goja.Runtime) func(goja
 			panic(r.NewTypeError("ttl must be 0 or more"))
 		}
 
-		n.localCache.Put(key, value.Export(), ttl)
+		v := value.Export()
+
+		switch v.(type) {
+		case string, int64, float64, bool:
+		default:
+			panic(r.NewTypeError("unsupported value type: must be string, numeric or boolean"))
```
syhpoon, line 8012: Is there a chance this would cause panics in existing modules that use types outside of this list?
I.e. would it make sense to issue a warning/error instead of panicking to prevent breaking backward compatibility?

syhpoon (approved):

server/runtime_javascript_nakama.go
sesposito: Goja types that are passed by reference (e.g.: a JS object) carry with them internally a reference to the original VM where they were instantiated, if these objects are manipulated concurrently it can cause the VM to panic.

PR #1150, opened 2023-12-05
Fix ptr slices assertions in js nk module arguments

sesposito:
Goja introduced the following change:
https://github.com/dop251/goja/commit/7749907a8a20719efef6c6fb4d39f2ba64f7557c JS arrays backed by a Go slice referenced by a pointer would be returned by .Export() as []S, but now are returned as *[]S which requires changing all the type assertions when dealing with these values.


ftkg (approved):

syhpoon (approved):

sesposito:
I added a commit as I had missed a few `.Export()` calls that needed patching.

syhpoon (approved):

PR #1153, opened 2023-12-13
Upgrade dependencies and Go docker images

sesposito:


data/modules/index.js
```diff
@@ -0,0 +1,538 @@
+"use strict";
```
syhpoon, line 1: was this file supposed to be a part of this commit? :thinking:

syhpoon (approved):

data/modules/index.js
sesposito: good catch, removed.

zyro (approved):

PR #1156, opened 2023-12-18
Improved matchmaking performance

Ratio2:
The order of combinations has changed slightly. Now the result is sorted according to the sorting of the input data.

Before these changes, matchmaking with override is very slow with more than 30 players:
players - time in seconds
10 - 1.1
20 - 1.9
22 - 2
24 - 2
26 - 3
28 - 8
30 - 28
32 - 118

### DIFF
--- a/vendor/github.com/mowshon/iterium/README.md
+++ b/vendor/github.com/mowshon/iterium/README.md
@@ -1,0 +1,709 @@
+<h1 id="iterium">🚀 Iterium - Generic Channel-based Iterators</h1>
+
+The **Iterium** package is a powerful toolkit for creating and manipulating generic iterators in Golang. Inspired by the popular Python **itertools** library, Iterium provides a variety of functions for working with iterators in different ways.
+
+Iterium is designed to be easy to use and flexible, with a clear and concise API that enables you to create iterators that meet your specific needs. Whether you're working with strings, arrays, slices, or any other data type. **Iterium** makes it easy to traverse, filter, and manipulate your data with ease.
+
+---------
+### Contents
+- [Decrypting the MD5 hash in Golang](https://github.com/mowshon/iterium#user-content-md5)
+    - [Benchmark](https://github.com/mowshon/iterium#user-content-benchmark)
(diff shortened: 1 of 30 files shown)


sesposito:
Hello @Ratio2, can you elaborate a bit more on what this PR is doing to improve performance?

Ratio2:
> Hello @Ratio2, can you elaborate a bit more on what this PR is doing to improve performance?

Input data:
`func combineIndexes(from []*MatchmakerIndex, min, max int)`
`len(from)` = 32
`min` = 2
`max` = 2

The function call duration is 118 seconds, since we have complexity `O(N*2^N)` from the input array size regardless of `min/max`. The loop `for combinationBits := 1; combinationBits < (1 << length); combinationBits++` consists of `2^32` iterations, which is a lot, and there is also an inner loop `for element := uint(0); element < length; element++`.

The main idea of ​​the change is to not iterate over impossible combinations in the loop. If `max` = 2, then we are not interested in combinations of 3 or more elements.

`iterium.Combinations` generates all possible combinations of `N` elements.
`rangeCombinations` generates all possible combinations of `1..N` elements. If the input is `[a, b, c]`, then the output is `[[a], [a, b], [a, c], [b], [b, c], [c]]`.

vendor/github.com/mowshon/iterium/README.md
```diff
@@ -0,0 +1,709 @@
...
+Before we move on to explore each iterator in particular, let me give you a small example of **decrypting an md5 hash in a few lines of code** using **Iterium**. Assume that our password consists only of **lower-case Latin letters** and we don't know exactly its length, but assume no more than 6 characters. 
+
+```golang
+// result of md5("qwerty") = d8578edf8458ce06fbc5bb76a58c5ca4
+passHash := "d8578edf8458ce06fbc5bb76a58c5ca4"
+
+for passLength := range Range(1, 7).Chan() {
+    fmt.Println("Password Length:", passLength)
+
+    // Merge a slide into a string.
```
rbrewer-echelon, line 48:
```suggestion
    // Merge a slice into a string.
```

PR #1159, opened 2024-01-18
Fix storage engine occ regression 

sesposito:
Fix an issue where storage object write with version would fail to return version rejected error if the same value is written consecutively using an invalid version.

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -655,7 +655,8 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 		)
 		(SELECT read, write, version, create_time, update_time, true AS update FROM upd)
 		UNION ALL
-		(SELECT read, write, version, create_time, update_time, false AS update FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))`
+		(SELECT read, write, version, create_time, update_time, false AS update FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))
+		LIMIT 1`
 
 		params = append(params, object.Version)
 
@@ -682,7 +683,8 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 		)
 		(SELECT read, write, version, create_time, update_time, true AS upsert FROM upd)
 		UNION ALL
-		(SELECT read, write, version, create_time, update_time, false AS upsert FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))`
+		(SELECT read, write, version, create_time, update_time, false AS upsert FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))
+		LIMIT 1`
 
 		// Outcomes:
 		// - Row is always returned, need to know if update happened, that WHERE matches
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -592,14 +592,16 @@ func storageWriteObjects(ctx context.Context, logger *zap.Logger, metrics Metric
 			return nil, err
 		}
 
-		// Indicates whether the operation inserted or updated the storage object value.
+		// Indicates whether the operation inserted or updated the storage object value in the db.
 		isUpsert := false
 		if upsert == 1 {
-			// Only if 1st query "order" value returns 1 the db op was an insert or update
+			// The db op was an insert or update only if "order" value returns 1 (see queries for clarification).
 			isUpsert = true
 		}
 
 		if !isUpsert {
+			// Conditions for successful update or insert were not met in the query, following checks are needed to disambiguate
+			// which error to return.
 			if !authoritativeWrite && resultWrite != 1 {
 				// - permission: non-authoritative write & original row write != 1
 				metrics.StorageWriteRejectCount(map[string]string{"collection": object.Collection, "reason": "permission"}, 1)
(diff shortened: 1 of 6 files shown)


redbaron (approved):

server/core_storage.go
```diff
@@ -647,24 +646,27 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
...
 		` + writeCheck + `
-			AND NOT (storage.version = $5 AND storage.read = $6 AND storage.write = $7) -- micro optimization: don't update row unnecessary
 			RETURNING read, write, version, create_time, update_time
 		)
-		(SELECT read, write, version, create_time, update_time from upd)
+		(SELECT read, write, version, create_time, update_time, true AS update FROM upd)
 		UNION ALL
-		(SELECT read, write, version, create_time, update_time FROM storage WHERE collection = $1 and key = $2 and user_id = $3)
+		(SELECT read, write, version, create_time, update_time, false AS update FROM storage WHERE collection = $1 and key = $2 and user_id = $3)
+		ORDER BY update DESC
```
redbaron, line 662: This variant of the query doesn't invoke unnecessary read of the same key and still produces correct result even if `UNION ALL` brances happen to be executed out of order:


```
WITH upd AS (
			UPDATE storage SET value = $4, version = $5, read = $6, write = $7, update_time = now()
			WHERE collection = $1 AND key = $2 AND user_id = $3 AND version = $8
		` + writeCheck + `
			RETURNING read, write, version, create_time, update_time
		)
(SELECT read, write, version, create_time, update_time, true AS update FROM upd WHERE EXISTS (SELECT FROM upd))
UNION ALL
(SELECT read, write, version, create_time, update_time, false AS update FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXIST (SELECT FROM upd)
LIMIT 1`
```

It works by making `UNION ALL` branches mutually exclusive with `EXISTS`/`NOT EXISTS` check on `upd` CTE. Data-modifying CTEs are documented to be executed only once:

> Data-modifying statements in WITH are executed exactly once, and always to completion, independently of whether the primary query reads all (or indeed any) of their output. 

So referencing CTE multiple times wont create multiple attempts to modify data and therefore cheap and safe.

server/core_storage.go
```diff
@@ -681,16 +682,15 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
...
 				UPDATE SET value = $4, version = $5, read = $6, write = $7, update_time = now()
 				WHERE TRUE` + writeCheck + `
-				AND NOT (storage.version = $5 AND storage.read = $6 AND storage.write = $7) -- micro optimization: don't update row unnecessary
 			RETURNING read, write, version, create_time, update_time
 		)
-		(SELECT read, write, version, create_time, update_time from upd)
+		(SELECT read, write, version, create_time, update_time, true AS upsert FROM upd WHERE EXISTS (SELECT 1 FROM upd))
 		UNION ALL
-		(SELECT read, write, version, create_time, update_time FROM storage WHERE collection = $1 and key = $2 and user_id = $3)
+		(SELECT read, write, version, create_time, update_time, false AS upsert FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))
```
redbaron, line 689: Doesn't need `1`, can be just a row with no columns: `SELECT FROM upd`

server/core_storage.go
```diff
@@ -681,16 +682,15 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 			ON CONFLICT (collection, key, user_id) DO
 				UPDATE SET value = $4, version = $5, read = $6, write = $7, update_time = now()
 				WHERE TRUE` + writeCheck + `
-				AND NOT (storage.version = $5 AND storage.read = $6 AND storage.write = $7) -- micro optimization: don't update row unnecessary
 			RETURNING read, write, version, create_time, update_time
 		)
-		(SELECT read, write, version, create_time, update_time from upd)
+		(SELECT read, write, version, create_time, update_time, true AS upsert FROM upd WHERE EXISTS (SELECT 1 FROM upd))
```
redbaron, line 687: Thinking about it, `WHERE EXISTS` here is redundant and can be removed:
- if `WHERE` is false, then select  also would not return row if written without `WHERE`.
- if `WHERE` is true, then select without `WHERE` always return row anyway.

server/core_storage.go
```diff
@@ -647,24 +646,26 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 		// allows us to fetch row state after update even if update itself fails WHERE
 		// condition.
 		// That is returned values are final state of the row regardless of UPDATE success
+		// Order by is necessary as UNION ALL may not guarantee ordering of results, but it is required
+		// to ensure that if update happens values from said update are returned.
+		// At most 2 rows are ever returned so ordering operation should be negligible.
```
redbaron, line 651: there is no sorting anymore

server/core_storage.go
sesposito: CRDB returns syntax without it.

server/core_storage.go
```diff
@@ -646,20 +646,16 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
...
 			WHERE collection = $1 AND key = $2 AND user_id = $3 AND version = $8
 		` + writeCheck + `
 			RETURNING read, write, version, create_time, update_time
 		)
-		(SELECT read, write, version, create_time, update_time, true AS update FROM upd WHERE EXISTS (SELECT 1 FROM upd))
+		(SELECT read, write, version, create_time, update_time, true AS update FROM upd)
 		UNION ALL
-		(SELECT read, write, version, create_time, update_time, false AS update FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))
-		LIMIT 1`
+		(SELECT read, write, version, create_time, update_time, false AS update FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))`
```
redbaron, line 658: `LIMIT 1` is still handy as it prevents whole branch of `UNION ALL` from exeucting if update succeeded. Without it it is always executed, but returns no rows on update success. End result is the same, but not executing should be ever so slightly cheaper on CPU .

You should be able to verify it with `EXPLAIN ANALYZE`, where `LIMIT 1` version shows  `(never executed)`   part of the plan.

redbaron (approved):

Issue #1160, opened 2024-01-24
ARM64 missing from docker registry


Darkbladecr:
As per issue #1138 there is actually no docker image for arm64/v8
```
! nakama The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

Now developers can build a local image and use it for development, but as @zyro mentioned:
> Note it is usually not recommended to build from source or build your own docker images, use existing binaries and images where possible.

Happy to put together a PR for the `docker buildx` commands but I don't actually see the `docker push` in the source code?

novabyte:
@Darkbladecr It's been on our backlog to introduce official ARM64 images with Docker buildx multi-arch container images. If you've got time it would be great to have the contribution. We store our Dockerfiles in this folder in the codebase:

https://github.com/heroiclabs/nakama/tree/master/build

Darkbladecr:
The binaries work fine when running the build scripts locally and the public releases. The only thing missing is to push the builds to docker. That’s why I was saying I don’t see any code in the repo that does this? It may be a CI pipeline that is not included in the repo? 

I have personally used docker buildx which is really easy to build multiple arch builds

otherwise since you have the pipeline already prepared with the standard docker build command we can just amend the manifest file to link to the multiple builds with one docker URI:
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

Darkbladecr:
Created a PR that updates the Dockerfile and created a script for multiarch builds for stable versions and prebuild:
#1161 

Tested on my `darwin/arm64` system, but will work on `linux/arm64` as well. If the CI environment is based on `linux/amd64` then the `CC and CCX` environmental variables will need to be updated

sesposito:
Resolved by: https://github.com/heroiclabs/nakama/pull/1316

Issue #1138, opened 2023-11-12
Build arm64 docker image get error


xinatcg:
Before you submit an issue we recommend you drop into the [community forum](https://forum.heroiclabs.com) and ask any questions you have or mention any problems you've had getting started with the server.

Please provide as much information as you can with this issue report. If you believe it may be an issue with one of the client libraries please report it on their [own trackers](https://github.com/heroiclabs?utf8=%E2%9C%93&q=nakama%20AND%20sdk&type=&language=).

## Description

Build arm64 docker image get error

```
docker build "$PWD" --platform "linux/arm64" --file ./Dockerfile.arm64 --build-arg commit="$(git rev-parse --short HEAD 2>/dev/null)" --build-arg version=3.19.0 -t local/nakama:3.19.0
```
 
 error:

```
8.692 gcc: error: unrecognized command-line option '-m64'
```
full

```
 => ERROR [builder 4/4] RUN git checkout --quiet "7d5051fb" &&     go build -o /go/build-out/nakama -trimpath -mod=vendor -gcflags "-trimpath $PWD" -asmflags "-trimpath $PWD" -ldflags "-s -w -X   34.1s
------
 > [builder 4/4] RUN git checkout --quiet "7d5051fb" &&     go build -o /go/build-out/nakama -trimpath -mod=vendor -gcflags "-trimpath $PWD" -asmflags "-trimpath $PWD" -ldflags "-s -w -X main.version=3.19.0 -X main.commitID=7d5051fb":
8.692 # runtime/cgo
8.692 gcc: error: unrecognized command-line option '-m64'
------
Dockerfile.arm64:34
--------------------
  33 |     WORKDIR /go/build/nakama
  34 | >>> RUN git checkout --quiet "$commit" && \
  35 | >>>     go build -o /go/build-out/nakama -trimpath -mod=vendor -gcflags "-trimpath $PWD" -asmflags "-trimpath $PWD" -ldflags "-s -w -X main.version=$version -X main.commitID=$commit"
  36 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c git checkout --quiet \"$commit\" &&     go build -o /go/build-out/nakama -trimpath -mod=vendor -gcflags \"-trimpath $PWD\" -asmflags \"-trimpath $PWD\" -ldflags \"-s -w -X main.version=$version -X main.commitID=$commit\"" did not complete successfully: exit code: 1

```

## Context

- [ ] Unity
- [ ] Unreal
[...]

PR #1161, opened 2024-01-25
feat(build): setup multiarch build for arm64

Darkbladecr:
According to issue #1160 #1138 

I do not know what the CI environment is for building the docker images. But I have updated the Dockerfile to be able to accommodate multiarch builds. Note my environment is `arm64` so I have to set up the docker for amd64, the converse may be true for your build CI environment.

As per the `multi_arch` script the docker buildx environment needs to be setup:

```bash
docker context create container
docker buildx create --use container
```

I assume the CI is a linux environment in which case buildx needs to be installed: https://github.com/docker/buildx?tab=readme-ov-file#linux-packages

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -15,26 +15,39 @@
 # docker build "$PWD" --build-arg commit="$(git rev-parse --short HEAD)" --build-arg version=v2.1.1 -t heroiclabs/nakama:2.1.1
 # docker build "$PWD" --build-arg commit="$(git rev-parse --short HEAD)" --build-arg version="$(git rev-parse --short HEAD)" -t heroiclabs/nakama-prerelease:"$(git rev-parse --short HEAD)"
 
-FROM golang:1.21.5-bookworm as builder
+FROM --platform=$BUILDPLATFORM golang:1.21.5-bookworm as builder
 
+ARG TARGETARCH
 ARG commit
 ARG version
 
 ENV GOOS linux
-ENV GOARCH amd64
+ENV GOARCH=$TARGETARCH
 ENV CGO_ENABLED 1
 
-RUN apt-get update && \
+RUN if [ "${TARGETARCH}" = "amd64" ]; then \
+    dpkg --add-architecture amd64; \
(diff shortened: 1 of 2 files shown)


mofirouz:
Thanks for this. Do you have a way to update the build README for binary releases too?

Darkbladecr:
> Thanks for this. Do you have a way to update the build README for binary releases too?

I haven't touched the binary workflow, so they should work as setup previously. I wanted to try and avoid any breaks with existing code or CI/CD. If you want me to clean everything up so that it works with this single Dockerfile, and update the readme for how to run it for any platform with `docker buildx` I can do that?

Darkbladecr:
Removed duplicate docker builds for both the main server and the pluginbuilder
Added info on how to use/setup in the [README.md](https://github.com/Darkbladecr/nakama/blob/master/build/README.md)

Updated the scripts to automatically pull the latest tag and info:

```bash
VER="$(git tag -l --sort=-creatordate | head -n 1)"
DOCKER_VER=heroiclabs/nakama:${VER:1}
echo Building $DOCKER_VER
docker buildx build --platform linux/amd64,linux/arm64 \
  --builder=container \
  --build-arg repo=https://github.com/heroiclabs/nakama.git \
  --build-arg commit="$(git rev-parse --short HEAD)" \
  --build-arg version=$VER \
  --tag $DOCKER_VER \
  --push .
```

Darkbladecr:
Anything else that needs to be done here @mofirouz ?

tobyw121:
hello , please have a look at my Github : https://github.com/tobyw121/nakama_docker_aarch64/tree/main

I have created a docker for a Nakama on aarch64 here .

sesposito:
Hello @Darkbladecr,

We've now resolved this issue on https://github.com/heroiclabs/nakama/pull/1316. I'd like to thank you for your contribution regardless, it was helpful to experiment on how to best achieve it, from my experimentation emulation turned out to be more reliable than cross-compilation so we went with that approach instead.

Best.

Issue #1164, opened 2024-02-02
Nakama-js can be timeout when pipeline.ProcessRequest return false


nguyentrannhan:
## Description
When pipeline.ProcessRequest return false, the nakama-js can be timeout because the IncommingLoop is break. To fix it, we add a flag for safely break the IncommingLoop

## Steps to Reproduce
1. Run nakama-js-webpack-example
2. Run joinChat with wrong input
3. Errors reported to client and in console log

## Expected Result
nakama-js should return with invalid input error not socket timeout

## Actual Result
Error is thrown with  socket timeout

## Context
- [ ] nakama-js

## Your Environment
- Nakama: 2.7.1
- Database: CockroachDB 
- Environment name and version: Development
- Operating System and version: Window 10/Ubuntu 20.04

ftkg:
I don't understand what the issue is supposed to be here and its attempted fix. If you are sending a malformed payload that cannot be processed, the socket will be closed as intended. What is the "wrong input" you are sending in `joinChat`?

zyro:
Discussed further [on the PR here](https://github.com/heroiclabs/nakama/pull/1165#issuecomment-2282810237).

PR #1165, opened 2024-02-02
when ProcessRequest is false and it sends error message to chan, the …

nguyentrannhan:
According to issue https://github.com/heroiclabs/nakama/issues/1164
nakama-js can be timeout because the IncomingLoop is break. We need to add a flag to break IncomingLoop safely

### DIFF
--- a/server/session_ws.go
+++ b/server/session_ws.go
@@ -171,6 +171,8 @@ func (s *sessionWS) Expiry() int64 {
 }
 
 func (s *sessionWS) Consume() {
+	var IsSafeBreakLoop bool = false
+
 	// Fire an event for session start.
 	if fn := s.runtime.EventSessionStart(); fn != nil {
 		fn(s.userID.String(), s.username.Load(), s.vars, s.expiry, s.id.String(), s.clientIP, s.clientPort, s.lang, time.Now().UTC().Unix())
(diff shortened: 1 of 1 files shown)


minhlucvan:
I am facing the same problem, please consider to merge this one

mofirouz:
The PR and code change is not descriptive enough.

- Can you please describe the actual problem you are facing? 
- Can you tell us what the gameplay mechanic you are building?
- What game engine (SDK) and what devices are you targeting? 

I ask these questions because we've never seen an issue in this code path.

nguyentrannhan:
I create this issue https://github.com/heroiclabs/nakama/issues/1164 for describe the problem what I am facing with. Please take a look.
Here is screen shot when I tested with joinChat function

The timeout message will appear after 10 seconds
image

zyro:
There is not enough information provided to understand why the server should process one additional message when an error is encountered with the socket connection.

If you can provide this additional required context then please feel free to re-open this, and we can evaluate it then. A complete minimal reproducible test case would also be very helpful. 🙏

nguyentrannhan:
This PR fix the issue about missing last error message due to the loop is closed. It is helpful to let client know the reason why server close the session.

PR #1166, opened 2024-02-02
Fix tournament active time calculation

sesposito:
Return correct active time when tournament duration collides with reset schedule.


mofirouz (approved):
Thanks dude - This must have been really annoying to find!

PR #1168, opened 2024-02-07
Add support to custom sorting in storage index search

hexun80149128:
https://forum.heroiclabs.com/t/does-the-query-in-storageindexlist-support-sorting-and-how/4948
 currently storage index listing does not support sorting the results.so add it。thanks

### DIFF
--- a/server/storage_index.go
+++ b/server/storage_index.go
@@ -20,7 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"reflect"
+	segment "github.com/blugelabs/bluge_segment_api"
 	"time"
 
 	"github.com/blugelabs/bluge"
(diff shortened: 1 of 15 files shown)


ftkg:
We probably also want to index the fields with the `Sortable()` property before doing requests with custom sorts, as specified in the [docs](https://blugelabs.com/bluge/sorting/).

hexun80149128:
RegisterStorageIndex add sortFields param.it must be in fields,and should be string or num. @ftkg

server/storage_index.go
```diff
@@ -483,6 +485,28 @@ func (si *LocalStorageIndex) mapIndexStorageFields(userID, collection, key, vers
 	rv.AddField(bluge.NewNumericField("read", float64(read)).StoreValue())
 	rv.AddField(bluge.NewNumericField("write", float64(write)).StoreValue())
 
+	if len(sortFilters) > 0 {
```
ftkg, line 488: This is duplicating the indexing of values that happens in `BlugeWalkDocument`. You could possibly set the `Sortable` property directly there, but keep in mind that other features also depend on this function.

hexun80149128: after BlugeWalkDocument set the Sortable property by sortFilters,and the sortFilters must start with "value.". I didn't modify BlugeWalkDocument as other features also depend on this function.

hexun80149128: @ftkg

ftkg: Thanks @hexun80149128 we'll evaluate how best to arrange this part of the code before merging it.

ftkg: @hexun80149128 can you please sign up the CLA so we can accept this PR later?

ftkg:
`nakama-common` PR: https://github.com/heroiclabs/nakama-common/pull/142

server/storage_index.go
hexun80149128: @ftkg done.

hexun80149128:
@ftkg can merge to master next version？

sesposito (approved):

sesposito:
@ftkg please add a changelog entry as well.

PR #1172, opened 2024-02-14
Reintroduce storage write query micro opt

sesposito:


server/core_storage.go
```diff
@@ -679,13 +681,16 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
...
 				UPDATE SET value = $4, version = $5, read = $6, write = $7, update_time = now()
 				WHERE TRUE` + writeCheck + `
+				AND NOT (storage.version = $5 AND storage.read = $6 AND storage.write = $7) -- micro optimization: don't update row unnecessarily
 			RETURNING read, write, version, create_time, update_time
 		)
-		(SELECT read, write, version, create_time, update_time, true AS upsert FROM upd)
+		(SELECT read, write, version, create_time, update_time, true AS upsert, false AS rejected FROM upd)
 		UNION ALL
-		(SELECT read, write, version, create_time, update_time, false AS upsert FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))
+		(SELECT read, write, version, create_time, update_time, false AS upsert, $8 != version AS rejected FROM storage WHERE collection = $1 and key = $2 and user_id = $3 AND NOT EXISTS (SELECT 1 FROM upd))
```
redbaron, line 689: non-OCC write (last write wins) can't fail with version rejected, so only way we can get here is on permission failure or microop kicking in, so `rejected` should always be false here

server/runtime_test.go
```diff
@@ -100,7 +100,9 @@ func runtimeWithModulesWithData(t *testing.T, modules map[string]string) (*Runti
 		leaderboardRankCache: lbRankCache,
 	}
 
-	rt, rtInfo, err := NewRuntime(ctx, logger, logger, db, protojsonMarshaler, protojsonUnmarshaler, cfg, "", nil, lbCache, lbRankCache, lbSched, nil, nil, nil, nil, nil, metrics, nil, &DummyMessageRouter{}, storageIdx)
+	tracker := &LocalTracker{}
```
redbaron, line 103: unrelated?

sesposito: tests were broken due to some changes that were introduced in another PR, unrelated but fixing them here

server/core_storage.go
```diff
@@ -651,11 +652,12 @@ func storagePrepBatch(batch *pgx.Batch, authoritativeWrite bool, op *StorageOpWr
 			UPDATE storage SET value = $4, version = $5, read = $6, write = $7, update_time = now()
 			WHERE collection = $1 AND key = $2 AND user_id = $3 AND version = $8
 		` + writeCheck + `
+			AND NOT (storage.version = $5 AND storage.read = $6 AND storage.write = $7) -- micro optimization: don't update row unnecessarily
```
redbaron, line 655: I don't think this optimization is applicable to OCC writes.  Assume db has version A and 2 concurrent transactions write version B , using A as OCC version condition. 

 If these 2 transactions are NOT overlapping in time, then second update will fail due to normal OCC version check.

If concurrent updates are overlapping, then one of them will block, waiting for another one to commit. Once it commits, it will recheck WHERE on the new row  and it will fail due to OCC check.

Only case this optimization can work is if versions A and B are the same version , that is app writes the same row as it read from DB before, which is very unlikely to be a common case and doesn't worth optimizing for

sesposito: This would indeed only apply if an object is re-written without applying any change to the last read version.

Whilst this shouldn't happen, I wouldn't be surprised if it does due to bad client logic at updating some shared object.

Even if unlikely, in the case it does happen it could save the db some work. Do you think it is not worth it regardless?

redbaron: without OCC support, whole change is probably just a single condition in non-OCC SQL query without even Go code changes.  Supporting OCC is more complicated with next to no benefit.

redbaron (approved):

PR #1174, opened 2024-02-20
Check Apple's LatestReceiptInfo for non-subscription purchases

### LINKED ISSUE #971: Apple incorrect IAP validation
the error `Subscription Receipt. Use the appropriate function instead.` is incorrect, you can no longer assume LatestReciept is empty, see https://developer.apple.com/forums/thread/63092

hhhapz:
Closes #971


hhhapz:
Hey, are there any further changes or edits I need to make? I would really like this to be in the next release :)

zyro (approved):

PR #1175, opened 2024-02-21
Add config option to enforce a single session token per user

### LINKED ISSUE #1014: [Feature Request] Logout or Invalidate Auth/Refresh token through Nakama backend
Using single_socket makes sure that there’s only one active sessions by disconnecting the previous login. The problem with this solution is that single_socket only disconnects socket, does not log out the user from that other device(previous login). This could pose as a security flaw as the auth token and refresh token are still valid.
There should be a way to invalidate these tokens(tokens from older login) or simply log out the older login maybe through a configuration along side single_socket on the backend.

Forum discussion:

My comment:
https://forum.heroiclabs.com/t/preventing-users-from-having-multiple-sessions/36/13

Chris Molozian's comment:
https://forum.heroiclabs.com/t/preventing-users-from-having-multiple-sessions/36/14

Chris has provided a solution in the above comment but that is still not secure because it is not Nakama authoritative which would be the most secure scenario.

ftkg:
Resolves #1014

### DIFF
--- a/server/session_cache.go
+++ b/server/session_cache.go
@@ -117,7 +117,7 @@ func (s *LocalSessionCache) IsValidSession(userID uuid.UUID, exp int64, tokenId 
 		s.RUnlock()
 		return true
 	}
-	if cache.lastInvalidation > 0 && exp-s.tokenExpirySec <= cache.lastInvalidation {
+	if cache.lastInvalidation > 0 && exp-s.tokenExpirySec < cache.lastInvalidation {
 		// The user has a full invalidation recorded, and this token's expiry indicates it was issued before this point.
 		s.RUnlock()
 		return false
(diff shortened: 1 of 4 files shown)


UmarBhatPlaystrom:
+1

waffillka:
+1

sesposito (changes requested):
@ftkg I believe that because `RemoveAll` invalidates all tokens emitted prior to the current time (with granularity within the same second), and because `IsValidSession` rejects tokens emitted within the same second too, this will also invalidate the token to be issued.

The new option should also be mentioned in the changelog.

server/session_cache.go
```diff
@@ -117,7 +117,7 @@ func (s *LocalSessionCache) IsValidSession(userID uuid.UUID, exp int64, tokenId
 		s.RUnlock()
 		return true
 	}
-	if cache.lastInvalidation > 0 && exp-s.tokenExpirySec <= cache.lastInvalidation {
+	if cache.lastInvalidation > 0 && exp-s.tokenExpirySec < cache.lastInvalidation {
```
ftkg, line 120: @sesposito Good catch, I've changed the validity check to only deny tokens issued strictly before the full invalidation. Do you see any problems with this?

sesposito (approved):

PR #1179, opened 2024-02-27
Add config option to enforce a single party per user socket

### LINKED ISSUE #1020: [Feature Request] Add config to disallow joining multiple parties
Add a global configuration such as `single_party` that if set, disallows from joining a party if already a member of one.

ftkg:
`single_party` should work like the existing `single_match` config, if set to `true` the user will leave the previous party it was a member of upon joining a new one.

Closes #1020


sesposito (approved):

sesposito:
@ftkg please add an entry to the changelog and we can merge this 👍

Issue #1123, opened 2023-10-31
Confirm Account Delete


delasource:
Please add a confirmation modal when deleting (recorded delete) a user account in Nakama Console, to prevent accidential button press.

mofirouz:
@delasource Yes that would be great; If you could create a PR, happy to accept contributions. Would be good to have it across multiple permanent-delete components in the UI. Thank you.

delasource:
I only know a little about Go and even less about Angular. So I'm the wrong person for this (:

xinatcg:
@mofirouz  PR submitted, hope this feature can be added in the near future.

sesposito:
Resolved in https://github.com/heroiclabs/nakama/pull/1197

PR #1197, opened 2024-04-10
feat: add delete confirmation dialog before real delete to prevent unexpected operation

xinatcg:
image

currently, all modal messages are the same, but the service supports custom titles and content for different usage.

https://github.com/heroiclabs/nakama/issues/1123


console/ui/src/app/shared/delete-confirm.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  providedIn: 'root'
+})
+export class DeleteConfirmService {
+
+  constructor(private modalService: NgbModal) {
+  }
+
+  openDeleteConfirmModal(confirmedCallback: () => void, title: string = '', message: string = ''): void {
+    const modalOptions: NgbModalOptions = {
+      backdrop: false,
```
sesposito, line 15: I suggest to add:
```
centered: true
```

sesposito (changes requested):
Thank you very much for your contribution! 
The PR looks good, however, if possible, could you also replace the usage of the `confirmDelete` modal in config and chatMessages with the same approach?
Before this can be merged, you'll also need to run `npm run-script build`

console/ui/src/app/shared/delete-confirm.service.ts
xinatcg: ok, np thanks

xinatcg:
> The PR looks good, however, if possible, could you also replace the usage of the `confirmDelete` modal in config and chatMessages with the same approach?

Ok, if so
> The PR looks good, however, if possible, could you also replace the usage of the confirmDelete modal in config and chatMessages with the same approach?

Should we leave them out and only keep the code for others without confirmation?

sesposito:
I'll approve without those changes, but it would be good to touch on them as part of this PR, if possible. I think those modals aren't being displayed correctly due to the upgrade to Angular 15 merge.

xinatcg:
> I'll approve without those changes, but it would be good to touch on them as part of this PR, if possible. I think those modals aren't being displayed correctly due to the upgrade to Angular 15 changes.

make some updates to support formGroup with 2 control in the confirm component

console/ui/src/app/config/config.component.ts
```diff
@@ -153,29 +155,27 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+            this.deleteError = '';
+            this.deleteSuccess = true;
+          }, err => {
+            this.deleting = false;
+            this.deleteError = err;
+          },
+        );
       },
+      this.confirmDeleteForm,
+      'Delete Project (NOT Configuration)' ,
```
sesposito, line 174:
```suggestion
      'Delete All Data' ,
```

console/ui/src/app/config/config.component.ts
```diff
@@ -153,29 +155,27 @@ export class ConfigComponent implements OnInit, OnDestroy {
...
+      'Are you sure you want to delete all data?'
```
sesposito, line 175:
```suggestion
     'Are you sure you want to delete all the database data?'
```
This should make it clearer, what do you think?

sesposito (approved):
Left a couple of suggestions but otherwise LGTM 👍

xinatcg:
> Left a couple of suggestions but otherwise LGTM 👍

update with 2 suggestions. thanks

sesposito:
Please rerun `npm run build` and we'll get this merged 👍

xinatcg:
> npm run-script build

done

xinatcg:
> Please rerun `npm run build` and we'll get this merged 👍

done

PR #1201, opened 2024-04-13
Fix event loop termination in statusRegistry

thesprockee:
Change `return` to `continue` in error handling to prevent the 
`statusRegistry` event loop from terminating.

### DIFF
--- a/server/status_registry.go
+++ b/server/status_registry.go
@@ -148,7 +148,7 @@ func NewLocalStatusRegistry(logger *zap.Logger, config Config, sessionRegistry S
 							payloadProtobuf, err = proto.Marshal(envelope)
 							if err != nil {
 								s.logger.Error("Could not marshal status event", zap.Error(err))
-								return
+								continue
 							}
 						}
 						err = session.SendBytes(payloadProtobuf, true)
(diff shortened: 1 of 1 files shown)


thesprockee:
This switches from `return` to `continue` in statusRegistry's event loop error handling to prevent `statusRegistry.eventsCh` queue from overflowing. 

This adjustment maintains the tracker event loop operation, which will terminate when the statusRegistry event queue overflows.

If the goroutine for the tracker event queue loop terminates, joinattempt markers will not be cleared causing match participants to be ejected from a match in ~15 seconds.

zyro:
@thesprockee I think strictly speaking this change would be correct, but I'd expect that error to only actually manifest in extreme scenarios like running out of memory etc - have you actually observed this error?

thesprockee:
Yes, the issue arose after I integrated a proprietary WebSocket protocol specific to a VR game into Nakama. Initially, I incorrectly used a "return" statement instead of "continue" upon detecting sessions using this proprietary protocol. Below is the original code segment, which helps illustrate why I initially overlooked the error. The repercussions became apparent only when the `statusRegistry.eventsCh` buffer exceeded 1024 events.


`status_registry.go`:

```golang
var err error
switch session.Format() {
case SessionFormatEvr:
	// Evr does not support status events.
	return
case SessionFormatProtobuf:
	if payloadProtobuf == nil {
		// Marshal the payload now that we know this format is needed.
		payloadProtobuf, err = proto.Marshal(envelope)
		if err != nil {
			s.logger.Error("Could not marshal status event", zap.Error(err))
			return
		}
	}
	err = session.SendBytes(payloadProtobuf, true)
case SessionFormatJson:
	fallthrough
```

PR #1209, opened 2024-05-06
Do not return value in console storage listing

sesposito:
The console storage list view has no use for the value, which for large stored values on low hardware deployments could cause an OOM.


zyro (approved):

zyro:
@sesposito Add a changelog entry too please. 🙏

server/console_storage.go
```diff
@@ -298,7 +298,7 @@ func (s *ConsoleServer) ListStorage(ctx context.Context, in *console.ListStorage
 		var createTime pgtype.Timestamptz
 		var updateTime pgtype.Timestamptz
 
-		if err := rows.Scan(&o.Collection, &o.Key, &o.UserId, &o.Value, &o.Version, &o.PermissionRead, &o.PermissionWrite, &createTime, &updateTime); err != nil {
+		if err := rows.Scan(&o.Collection, &o.Key, &o.UserId, &o.Version, &o.PermissionRead, &o.PermissionWrite, &createTime, &updateTime); err != nil {
```
flaviofernandes004, line 301: 💅  I think `console.proto` should be updated to reflect this change. Either removing the field since it only powers the UI Console AFAIK, or by marking it was deprecated.

PR #1210, opened 2024-05-07
server: update log level to error when checking apple token validity

dslizardo:
I'm seeing errors related to this check. In our deployment, log level was set to error. I think it would be nice to see the error from this method.


sesposito:
@dslizardo as discussed elsewhere we've opened an internal ticket to instead send a more informational error message back to the client on what caused a token to be rejected from the provider.

Closing this PR.

PR #1213, opened 2024-05-13
Add shutdown hook function support in the runtimes.

sesposito:
nakama-common PR: https://github.com/heroiclabs/nakama-common/pull/146


syhpoon (approved):
Would it make sense to add some tests for this new feature as well?

PR #1219, opened 2024-05-28
Upgrade gopher-lua to v1.1.1

sesposito:


syhpoon (commented):
hm, what's up with all this inlining business here? 
Why are we doing compiler's job? :thinking:

zyro (approved):

sesposito:
Will open new PR to upgrade to gopher-lua v1.1.2 instead.

PR #1224, opened 2024-06-04
avoid panicking if TagRPC wasn't called for any reason

syhpoon:


lugehorsam (approved):

intinig:
For more context I think this fix is a necessary mitigation related to a potential bug in go-grpc: https://github.com/grpc/grpc-go/issues/6928.

zyro (approved):

PR #1229, opened 2024-06-06
Add friends of friends api and runtime functions

sesposito:
nakama-common PR: https://github.com/heroiclabs/nakama-common/pull/148


server/api_friend.go
```diff
@@ -89,6 +89,61 @@ func (s *ApiServer) ListFriends(ctx context.Context, in *api.ListFriendsRequest)
...
+		err := traceApiBefore(ctx, s.logger, s.metrics, ctx.Value(ctxFullMethodKey{}).(string), beforeFn)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	limit := 10
+	if in.GetLimit() != nil {
+		if in.GetLimit().Value < 1 || in.GetLimit().Value > 100 {
+			return nil, status.Error(codes.InvalidArgument, "Invalid limit - limit must be between 1 and 1000.")
```
syhpoon, line 121: is it 100 or 1000? :thinking:

syhpoon (approved):
I know it's a tight timeline, but do you think there's a chance you could also squeeze in at least some basic tests of this new functionality?
If not, we can do this afterwards.

server/api_friend.go
zyro: For direct client operations it should be 100 maximum. Runtime calls I can see a case for allowing 1000 or even more as maximum.

zyro (approved):

server/api_friend.go
syhpoon: my point is that the returned error is misleading (containing 1000)

sesposito: I've addressed the comment and increased the limit to 1000 for calls from the runtimes.

PR #1232, opened 2024-06-17
Change rank cache delete condition

antontidev:
For case when leaderboard is created without reset schedule and you want to delete leaderboard you also need to remove rank cache

If you don't delete rank cache after leaderboard deletion and create leaderboard with same id you will get wrong rank values for owner


sesposito:
Hello @antontidev, thank you for your contribution! Please sign the CLA so we can accept it.

sesposito (approved):

PR #1233, opened 2024-06-21
HandleRPC does not panic anymore on an untagged gRPC call

intinig:
The fix applied to v3.22.0 does not completely solve the problem of untagged gRPCs, with the panic now moved to a nil dereference to `m.Metrics.CountUntaggedGrpcStatsCalls(1)`.

I applied a simple fix by wrapping it in a double check for non-nil, this makes it work again for us.


syhpoon:
Thanks @intinig !
This PR (https://github.com/heroiclabs/nakama/pull/1234) should solve the root cause of this problem.

intinig:
thx

PR #1245, opened 2024-07-09
inject go modules directly for easier debugging.

mertcelen:
Debugging with go modules are quite difficult, so with that pull request nakama' startup sequence moved to package based system. It shouldn't effect regular flow of nakama instance by default, but it will allow injecting go modules directly to core instance.

If you want to test this, clone this source branch. lets say path is "/Users/mertcelen/new-nakama"

example code should like like this

```
package main

import (
	"context"
	"database/sql"
	nakamaRuntime "github.com/heroiclabs/nakama-common/runtime"
	nakamaInstance "github.com/heroiclabs/nakama/v3/cmd/instance"
	nakamaServer "github.com/heroiclabs/nakama/v3/server"
)

func main() {
	mainInstance := nakamaInstance.New()
	mainInstance.AddGoModule(Test)
	mainInstance.Start()
}

func Test(ctx context.Context, logger nakamaRuntime.Logger, db *sql.DB, module *nakamaServer.RuntimeGoNakamaModule, initializer *nakamaServer.RuntimeGoInitializer) error {
	return nil
}
```

and you should add this line to go.mod file below import to be able to test the new code
```
replace github.com/heroiclabs/nakama/v3 => /Users/mertcelen/new-nakama
```

final should like this
```
module nakamaDebugger

go 1.22

require github.com/heroiclabs/nakama/v3 v3.22.0

replace github.com/heroiclabs/nakama/v3 => /Users/mertcelen/new-nakama
require (
    ...
```


zyro:
We recently revamped our [Debugging with Delve](https://heroiclabs.com/docs/nakama/guides/server-framework/debugging-with-delve/) documentation, that is our official recommendation for debugging Go runtime code or its integration with Nakama's internals.

Re-flowing the integration as you suggest might be possible, but it should not be necessary. 🙇

PR #1247, opened 2024-07-12
Add notification aux functions

sesposito:
NK-583
NK-584

https://github.com/heroiclabs/nakama-common/pull/153

### DIFF
--- a/server/core_notification.go
+++ b/server/core_notification.go
@@ -20,7 +20,10 @@ import (
 	"database/sql"
 	"encoding/base64"
 	"encoding/gob"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"github.com/heroiclabs/nakama-common/runtime"
 	"time"
 
 	"github.com/gofrs/uuid/v5"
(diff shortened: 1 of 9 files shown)


server/core_notification.go
```diff
@@ -322,3 +325,69 @@ SELECT
...
+		}
+		notifications = append(notifications, no)
+	}
+
+	return notifications, nil
+}
+
+func NotificationsDeleteId(ctx context.Context, logger *zap.Logger, db *sql.DB, ids ...string) error {
+	if len(ids) == 0 {
+		return errors.New("expects at least one id")
```
zyro, line 378: I suggest we just no-op this instead.

server/core_notification.go
```diff
@@ -322,3 +325,69 @@ SELECT
 
 	return nil
 }
+
+func NotificationsGetId(ctx context.Context, logger *zap.Logger, db *sql.DB, ids ...string) ([]*runtime.Notification, error) {
```
zyro, line 329: We should include the user ID and caller ID parameters so we can optionally enforce access controls.

server/core_notification.go
```diff
@@ -322,3 +325,69 @@ SELECT
...
+		if no.Sender == uuid.Nil.String() {
+			no.Sender = ""
+		}
+		notifications = append(notifications, no)
+	}
+
+	return notifications, nil
+}
+
+func NotificationsDeleteId(ctx context.Context, logger *zap.Logger, db *sql.DB, ids ...string) error {
```
zyro, line 376: Same as above, user ID and caller ID params.

zyro (changes requested):

zyro (approved):

PR #1252, opened 2024-07-26
Allow Public RPC Registration

thegoldenmule:
Full description of the problem is here: https://github.com/heroiclabs/nakama/issues/537

#### Proposed Solution:

I added a new method, `RegisterPublicRpc` (and `register_public_rpc` for Lua). This method allows you to register RPCs that do not require authentication.

An end-to-end test is included in `runtime_test.go`, named `TestRuntimeRegisterPublicRPCWithPayloadEndToEnd`. This successfully calls a public RPC without any HTTP key.

Most of the work is in `api_rpc.go`, and there are a few specific details to note:

1. Previously, metrics were being gathered for RPC calls without an RPC id or with a bad RPC id. This check was moved forward to simplify some logic and so metrics are no longer gathered in those circumstances.

2. The simplest method of bypassing auth was simply to branch. This means that the userId and username will be _empty_ for public RPC handlers, regardless of whether or not they are authenticated. There are a couple other cases in which the handler would have been called with uid and username, but I specifically stripped that data out to guarantee that those values will always be empty. This seems like desired behavior, but obviously up to you.


thegoldenmule:
@novabyte Hi there, any indication on how this is being received internally? It'd be great to understand if this is something that your team thinks could be accepted.

novabyte:
@thegoldenmule Thanks for the ping. I'll make sure its added to our next internal Nakama meeting so we can make suggestions. /cc @zyro

waldobronchart:
Friendly ping on this, I'm waiting on this feature too.

thegoldenmule:
@novabyte Any news?

MaksArh:
friendly ping, any news?

thegoldenmule:
Any news @novabyte? Or perhaps from @zyro the hero?

zyro:
We're looking to include this change in a Nakama release by end of this month. 👍

MaksArh:
what about anonym auth with device? i think its looks like a solution now

thegoldenmule:
w0000000000 let's GO

thegoldenmule:
> what about anonym auth with device? i think its looks like a solution now

@MaksArh Unless I'm misunderstanding you, that is already supported. See https://heroiclabs.com/docs/nakama/concepts/authentication/#device

MaksArh:
> > what about anonym auth with device? i think its looks like a solution now
> 
> @MaksArh Unless I'm misunderstanding you, that is already supported. See https://heroiclabs.com/docs/nakama/concepts/authentication/#device

I mean, if you need to use rpc before authentication (for example, 2fa), you can use a temporary (anonymous) login via the device

thegoldenmule:
Oh I see what you mean. Yah, potentially possible but obviously not ideal for a number of reasons. We don't want to have to create accounts to get public data, device login or not. It would be very tricky to get right, especially for users that actually do use device login, but for whom we want to cohort for login itself.

zyro:
See [this issue comment](https://github.com/heroiclabs/nakama/issues/537#issuecomment-2447469644) for updates. For the moment an alternative approach has been necessary to handle all requirements.

thegoldenmule:
@zyro @novabyte Really appreciate the movement on this, thank you for a comprehensive approach. We are looking forward to digging in.

I hesitated to say anything, especially publicly, but I really like Nakama and want you guys to succeed, so I want to offer some critical feedback that is hopefully received constructively. If I'm honest, this process has reflected very poorly on Heroic Labs, displaying not an open-source product but an "open-source in name only" product.

If you trace back through the feature request and PR, this is what happened:

1. We request a feature, expounding on use cases.
2. We are told it's a great idea, being discussed internally.
3. Nothing happens for months.
4. We ask if we can help out and are told "We're happy to review a pull request and consider it for inclusion in a future release."
5. We do a lot of work, making a PR that fits the ask.
6. No feedback.
7. Eventually, we get a "We're looking to include this change in a Nakama release by end of this month. 👍". Hey we're excited, who wouldn't be.
8. A full three months afterward the PR is closed and there is a completely different, unmentioned, and undocumented approach. There was no collaboration, so almost by definition you have no idea if it "handles all requirements".

I would ask you to consider that, given this kind of process working with the open-source community, why the heck anyone would consider contributing to this project. I get that some of your code is public, which is great (and kudos to your team) because then idiots like me can better understand the product. However, having a public repository on GitHub is _lightyears away_ from having an open-source product. This broken process in between of actually working with the community is the gaping chasm.

zyro:
@thegoldenmule I appreciate the feedback and your directness. We'll take this on board as we continue to improve our processes here and on our various other repos. 🙇

PR #1253, opened 2024-07-28
Removed some unnecessary fallthrough statements

toqueteos:
There are still 29 non-vendor `fallthrough` usages left, which make sense (most of them are specific `case`s that should behave as the `default` case). So no need to touch those.

The most weird one is the one on `iap/iap.go` which was actually not doing anything at all (besides listing two possible statuses?).


sesposito (approved):

toqueteos:
Gentle ping @sesposito :wave: !

Do you need me to provide/change something else or is this good to go?

sesposito:
Hey @toqueteos, it's good to go, I'll merge it, thanks!

PR #1262, opened 2024-08-09
Remove usages of io/ioutil

toqueteos:
Package [io/ioutil](https://pkg.go.dev/io/ioutil) deprecated [since Go 1.16](https://go.dev/doc/go1.16#ioutil)

There are some `io/ioutil` usages left in `internal/gopher-lua` which I've sent upstream (plus some test speedups) https://github.com/yuin/gopher-lua/pull/498


sesposito (approved):
Thank you @toqueteos

PR #1342, opened 2025-03-19
Upgrade all dependencies

sesposito:
Resolves NK-650 NK-648


sesposito:
Resolves NK-650 NK-648

zyro (approved):

PR #2330, opened 2025-05-31
Allow optional count parameter in nk.TournamentRecordWrite to update 'num_score' field

HanifNaOH:


novabyte:
@HanifNaOH Thanks for the contribution. 🙏 

It would be great to include this into Nakama, can you provide a small description of the feature improvement? What type of game have you added this for?

PR #2331, opened 2025-06-05
Fix Groups annotation

DannyIsYog:


server/runtime_go_nakama.go
```diff
@@ -3901,7 +3901,7 @@ func (n *RuntimeGoNakamaModule) GroupUsersList(ctx context.Context, groupID stri
 // @param ctx(type=context.Context) The context object represents information about the server and requester.
 // @param name(type=string, optional=true) Search for groups that contain this value in their name. Cannot be combined with any other filter.
 // @param langTag(type=string, optional=true) Filter based upon the entered language tag.
-// @param members(type=int, optional=true) Search by number of group members.
+// @param members(type=int, optional=true) Search groups with an equal or lower number of members.
```
sesposito:
```suggestion
// @param members(type=int, optional=true) Search groups with an equal or lower number of members in descending order.
```

sesposito: small tweak that just occurred to me 🙏

sesposito (changes requested):

sesposito (approved):

PR #2333, opened 2025-06-06
NK-589 Add Party Search support

sesposito:
Allow to set a party label on creation.
Add PartyUpdate socket function to update the label.
Add PartyList REST and runtime API.


server/party_registry.go
```diff
@@ -51,34 +73,146 @@ type LocalPartyRegistry struct {
...
+			return nil, fmt.Errorf("failed to unmarshal party label: %s", err.Error())
+		}
+	}
+
 	partyHandler := NewPartyHandler(p.logger, p, p.matchmaker, p.tracker, p.streamManager, p.router, id, p.node, open, maxSize, presence)
 
 	p.parties.Store(id, partyHandler)
 
-	return partyHandler
+	if labelMap != nil {
```
zyro: I think we should index parties even without a label, so they're still searchable by like `*` query.

server/party_registry.go
```diff
@@ -214,3 +353,258 @@ func (p *LocalPartyRegistry) PartyDataSend(ctx context.Context, id uuid.UUID, no
...
+	if err := ph.Update(sessionID, fromNode, label, open); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *LocalPartyRegistry) LabelUpdate(id uuid.UUID, node, label string, open bool, maxSize int, createTime time.Time) error {
+	idStr := fmt.Sprintf("%v.%v", id.String(), node)
+	if label == "" {
```
zyro, line 382: Need to handle the case where the label is set back to empty {} but stays in the index. The index entry should only be fully deleted when the party is deleted/closed.

server/runtime_javascript_nakama.go
```diff
@@ -446,6 +449,64 @@ func (n *RuntimeJavascriptNakamaModule) storageIndexList(r *goja.Runtime) func(g
...
+// @param limit(type=number, optional=true, default=10) The maximum number of parties to list.
+// @param open(type=bool, optional=true, default=null) Filter open or closed parties. If null, both open and closed parties are returned.
+// @param query(type=string, optional=true) Additional query parameters to shortlist parties.
+// @param cursor(type=string, optional=true) A cursor to fetch the next page of results.
+// @return parties(nkruntime.Party[]) A list of parties matching the filtering criteria.
+// @return error(error) An optional error value if an error occurred.
+func (n *RuntimeJavascriptNakamaModule) partiesList(r *goja.Runtime) func(goja.FunctionCall) goja.Value {
+	return func(f goja.FunctionCall) goja.Value {
+		limit := 10
+		if f.Argument(0) != goja.Undefined() && f.Argument(0) != goja.Null() {
```
zyro, line 463: Do we need to check limit bounds 1-100 here?

server/runtime_lua_nakama.go
```diff
@@ -10332,6 +10335,61 @@ func (n *RuntimeLuaNakamaModule) friendsBlock(l *lua.LState) int {
...
+// @param limit(type=number, optional=true, default=10) The maximum number of parties to list.
+// @param open(type=bool, optional=true, default=null) Filter open or closed parties. If null, both open and closed parties are returned.
+// @param query(type=string, optional=true) Additional query parameters to shortlist parties.
+// @param cursor(type=string, optional=true) A cursor to fetch the next page of results.
+// @return parties(table) A list of parties matching the filtering criteria.
+// @return cursor(string) A cursor to fetch the next page of results.
+// @return error(error) An optional error value if an error occurred.
+func (n *RuntimeLuaNakamaModule) partiesList(l *lua.LState) int {
+	// Parse limit.
+	limit := l.OptInt(1, 10)
```
zyro, line 10349: Check limit 1-100 here?

server/party_registry.go
sesposito: I thought we may want to prevent some parties from being listed, keeping them private, wdyt?

zyro (approved):

server/party_registry.go
sesposito: Added "hidden" option for parties to filter them from client listing results.

PR #2348, opened 2025-07-02
Update tournament size on record deletion.

sesposito:


zyro (commented):
Do we need to check if a deletion was actually done, and only decrement size in that case?

sesposito:
Good catch, added required check.

zyro (approved):

PR #2365, opened 2025-08-13
Gracefully handle socket closure.

sesposito:


server/session_ws.go
```diff
@@ -514,12 +517,35 @@ func (s *sessionWS) Close(msg string, reason runtime.PresenceReason, envelopes .
...
+			s.logger.Warn("Failed to set read deadline", zap.Error(err))
+		} else {
+			for {
+				msgType, _, readErr := s.conn.ReadMessage()
+				if readErr != nil {
+					// If any error occurs, close message likely won't be delivered, just close the socket.
+					break
+				}
+				if msgType == websocket.CloseMessage {
+					// We only care about the close message at this point, if the server initiated a close then something went wrong.
```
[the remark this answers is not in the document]
sesposito: The point is that the server only initiates a close on its own if something went wrong, otherwise closure would be caused by network issues or client closing it gracefully.

server/session_ws.go
```diff
@@ -514,12 +517,39 @@ func (s *sessionWS) Close(msg string, reason runtime.PresenceReason, envelopes .
 		// This may not be possible if the socket was already fully closed by an error.
 		s.logger.Debug("Could not send close message", zap.Error(err))
 	}
+	if msg != "" {
+		// Server initiated close, await for client close response.
+		t := time.NewTimer(10 * time.Second)
```
zyro: Does `conn.SetReadDeadline` not have the same effect?

sesposito: This should be a safeguard in case the client misbehaves and keeps sending non-closure messages

zyro (approved):

PR #2366, opened 2025-08-13
Rerun generate protocols step

RicardRC:


sesposito:
Protobuf stubs have since been updated, closing.

PR #2378, opened 2025-08-27
Satori client fixes and improvements.

sesposito:
Add missing fields to returned flag listing entries. 
Make cache eviction more aggressive to prevent constant memory growth in some workloads.
Trigger GH actions on publish instead of create.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/gorilla/websocket v1.5.3
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.27.2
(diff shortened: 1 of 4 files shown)


martinfidel-bandai:
This is great! It is exactly what I was looking for 🎉 

Just one small comment, is the replace in go mod intended to be committed? seems like normally a development environment related thing.

zyro (approved):

zyro:
> This is great! It is exactly what I was looking for 🎉
> 
> Just one small comment, is the replace in go mod intended to be committed? seems like normally a development environment related thing.

Good catch @martinfidel-bandai! It should be stripped out before the PR lands. 👍

zyro (approved):

PR #2397, opened 2025-10-01
Add fine grained ACL for Console users.

sesposito:


console/acl/acl.go
```diff
@@ -0,0 +1,392 @@
...
+		if permissions.Write {
+			acc = acc.Compose(NewPermissionFromString(resource, PermissionWrite))
+		}
+
+		if permissions.Delete {
+			acc = acc.Compose(NewPermissionFromString(resource, PermissionDelete))
+		}
+	}
+
+	return acc
```
flaviofernandes004, line 332: Remember to check the equivalence for the permissions received with the admin. The admin is someone with the set of bits all with 1. Unless you have all bits reserved for all resources, you will receive a subset bits only for declared resources and that number won't match the array with all bits with 1 but it will be equivalent.

PR #2404, opened 2025-10-23
Fix missing variable return in google validate subscription

Kyle-CW:
In the `GetSubscriptionV2Google` function, the `buf` variable is never returned but it is relied upon by the calling `ValidateSubscriptionGoogle` 
https://github.com/heroiclabs/nakama/blob/e2b7993b6ca1149f24d44b8a111d0e1c920e3d49/server/core_subscription.go#L363-L364

via 

https://github.com/heroiclabs/nakama/blob/e2b7993b6ca1149f24d44b8a111d0e1c920e3d49/iap/iap.go#L524-L529

This raw response should then be inserted into the stored subscription model
https://github.com/heroiclabs/nakama/blob/e2b7993b6ca1149f24d44b8a111d0e1c920e3d49/server/core_subscription.go#L383-L392

but this value is always nil currently.

### DIFF
--- a/iap/iap.go
+++ b/iap/iap.go
@@ -639,7 +639,7 @@ func GetSubscriptionV2Google(ctx context.Context, httpc *http.Client, clientEmai
 			return nil, nil, err
 		}
 
-		return out, nil, nil
+		return out, buf, nil
 	default:
 		return nil, nil, &ValidationError{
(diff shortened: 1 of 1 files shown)


sesposito (approved):
Nice one @Kyle-CW!

PR #2415, opened 2025-11-11
Add runtime API to generate secure random bytes via crypto/rand.

caprica:
Add a new function for Javascript and Lua runtimes to generate one or more cryptographically secure random numbers using crypto/rand.

This is used for example to generate random numbers that can be used to perform a secure shuffle, which is necessary for certain types of games where the normal random number generator (e.g. Math.random) should not be used.

The number of requested bytes was arbitrarily range-checked to be 1-1000, to at least put some limit on it.

Tested and working in a private Nakama build.

Some discussion here: https://forum.heroiclabs.com/t/cryptographically-secure-random-numbers-in-typescript-module

### DIFF
--- a/server/runtime_lua_nakama.go
+++ b/server/runtime_lua_nakama.go
@@ -10447,7 +10447,7 @@ func (n *RuntimeLuaNakamaModule) secureRandomBytes(l *lua.LState) int {
 	count := l.CheckInt(1)
 
 	if count < 1 || count > 1000 {
-		l.RaiseError("count must be 1-1000")
+		l.ArgError(1, "count must be 1-1000")
 		return 0
 	}
 


server/runtime_lua_nakama.go
```diff
@@ -10437,6 +10438,37 @@ func (n *RuntimeLuaNakamaModule) partyList(l *lua.LState) int {
...
+// @group utils
+// @summary Generate cryptographically secure random bytes.
+// @param count(type=int) The number of bytes to generate, from 1 to 1000.
+// @return bytes(string) The cryptograpically secure bytes.
+// @return error(error) An optional error value if an error occurred.
+func (n *RuntimeLuaNakamaModule) secureRandomBytes(l *lua.LState) int {
+	count := l.CheckInt(1)
+
+	if count < 1 || count > 1000 {
+		l.RaiseError("count must be 1-1000")
```
zyro: I think `ArgError` is more appropriate here?

PR #2419, opened 2025-11-19
Additional cache modes for Satori client.

zyro:


internal/satori/satori.go
```diff
@@ -85,10 +90,25 @@ func NewSatoriClient(ctx context.Context, logger *zap.Logger, satoriUrl, apiKeyN
...
-		flagsOverridesCache:  newSatoriCache[flagOverridesCacheEntry](ctx, cacheEnabled),
-		liveEventsCache:      newSatoriCache[*runtime.LiveEvent](ctx, cacheEnabled),
-		experimentsCache:     newSatoriCache[*runtime.Experiment](ctx, cacheEnabled),
+		flagsCache:           newSatoriContextCache[flagCacheEntry](ctx, cacheEnabled),
+		flagsOverridesCache:  newSatoriContextCache[flagOverridesCacheEntry](ctx, cacheEnabled),
+		liveEventsCache:      newSatoriContextCache[*runtime.LiveEvent](ctx, cacheEnabled),
+		experimentsCache:     newSatoriContextCache[*runtime.Experiment](ctx, cacheEnabled),
+	}
+
+	switch cacheMode {
```
sesposito: Should we make this an enum to have more explicit values?

internal/satori/satori.go
```diff
@@ -85,10 +90,25 @@ func NewSatoriClient(ctx context.Context, logger *zap.Logger, satoriUrl, apiKeyN
 		cacheEnabled:         cacheEnabled,
```
sesposito, line 90: Should we consider removing from the client struct and just allow the cache to behave accordingly?

sesposito (approved):

internal/satori/satori.go
zyro: For the moment it is still used for the properties, so until that's refactored we need it here too.

PR #2440, opened 2025-12-12
Audit log improvements

sesposito:
Obfuscate console signing key from config page.

### DIFF
--- a/server/console_user.go
+++ b/server/console_user.go
@@ -25,7 +25,6 @@ import (
 	"regexp"
 	"strings"
 	"time"
-	"unicode"
 
 	"github.com/gofrs/uuid/v5"
 	"github.com/golang-jwt/jwt/v5"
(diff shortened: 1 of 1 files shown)


server/console_audit.go
```diff
@@ -158,6 +159,28 @@ func consoleAuditLogInterceptor(logger *zap.Logger, db *sql.DB) func(context.Con
...
+				pwdResetReq, ok := req.(*console.AuthenticatePasswordChangeRequest)
+				if ok {
+					tokens := strings.Split(pwdResetReq.GetCode(), ".")
+					if len(tokens) == 3 {
+						decodeString, err := base64.RawURLEncoding.DecodeString(tokens[1])
+						if err != nil {
+							break
+						}
+						var claims map[string]any
+						if err = json.Unmarshal(decodeString, &claims); err != nil {
```
sesposito, line 172: Token does not require validation at this stage because it has already been validated by the request handler. If validation would have failed, we'd have skipped insertion altogether.

server/console_authenticate.go
```diff
@@ -112,7 +113,8 @@ func (s *ConsoleServer) Authenticate(ctx context.Context, in *console.Authentica
 				uname = in.Username
 				userId = uuid.Nil
 
-				hashedPassword, err := bcrypt.GenerateFromPassword([]byte(in.Password), bcrypt.DefaultCost)
+				hashCost := math.Max(float64(bcrypt.DefaultCost), 13)
```
zyro: Maybe move this to a global var rather than create if every time?

zyro (approved):

PR #2457, opened 2026-01-30
Fix Lua RNG seeding

VBorisof:
Fixes [Linear: math.randomseed() in Lua runtime is broken since 3.27.0](https://linear.app/heroiclabs/issue/NK-1705/bug-report-mathrandomseed-in-lua-runtime-is-broken-since-3270).

Introduces `rand.Rand` instance and seeds that, as [rand.Seed() is deprecated.](https://pkg.go.dev/math/rand#Seed)

### DIFF
--- a/internal/gopher-lua/mathlib.go
+++ b/internal/gopher-lua/mathlib.go
@@ -8,32 +8,28 @@ import (
 )
 
 type LuaRand struct {
-	lock sync.Mutex
+	sync.Mutex
 	rand *rand.Rand
 }
 
 func (r *LuaRand) Float64() (out float64) {
-	r.lock.Lock()
+	r.Lock()
 	out = r.rand.Float64()
-	r.lock.Unlock()
+	r.Unlock()
 	return
 }
 
 func (r *LuaRand) Intn(n int) (out int) {
-	r.lock.Lock()
+	r.Lock()
 	out = r.rand.Intn(n)
-	r.lock.Unlock()
+	r.Unlock()
 	return
 }
 
-var random *LuaRand
-
-func init() {
-	random = &LuaRand{
-		rand: rand.New(rand.NewSource(
-			time.Now().UTC().UnixMilli(),
-		)),
-	}
+var random *LuaRand = &LuaRand{
+	rand: rand.New(rand.NewSource(
+		time.Now().UTC().UnixMilli(),
+	)),
 }
 
 func OpenMath(L *LState) int {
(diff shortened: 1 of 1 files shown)


novabyte:
@VBorisof It might be worth a sync of the `gopher-lua` repo which we mirror as an internal dependency for any fixes since our last sync at this commit: https://github.com/heroiclabs/nakama/commit/4bccf1d6ec8ff254b4a72de3f84dcb977c7419fc

internal/gopher-lua/mathlib.go
```diff
@@ -3,8 +3,39 @@ package lua
 import (
 	"math"
 	"math/rand"
+	"sync"
+	"time"
 )
 
+type LuaRand struct {
+	lock sync.Mutex
```
sesposito:
```suggestion
	sync.RWMutex
```
You should use a RWMutex here to allow concurrent readers.

nit: You can omit naming the lock, this will effectively embed the lock into your declared struct, allowing you to call: r.Lock/Unlock() below.

internal/gopher-lua/mathlib.go
```diff
@@ -3,8 +3,39 @@ package lua
...
+	rand *rand.Rand
+}
+
+func (r *LuaRand) Float64() (out float64) {
+	r.lock.Lock()
```
sesposito: use `RLock()/RUnlock()` here so that there can be multiple concurrent readers, the underlying `rand.Float64()` function is concurrent safe as per the rand package documentation.

internal/gopher-lua/mathlib.go
```diff
@@ -3,8 +3,39 @@ package lua
...
+	out = r.rand.Float64()
+	r.lock.Unlock()
+	return
+}
+
+func (r *LuaRand) Intn(n int) (out int) {
+	r.lock.Lock()
```
sesposito: ditto

internal/gopher-lua/mathlib.go
```diff
@@ -3,8 +3,39 @@ package lua
...
+	out = r.rand.Intn(n)
+	r.lock.Unlock()
+	return
+}
+
+var random *LuaRand
+
+func init() {
```
sesposito: nit: I know I suggested the use of `init` myself, but perhaps it's simpler to just assign the variable on the global declaration instead.

internal/gopher-lua/mathlib.go
redbaron: from docs

> Random numbers are generated by a [Source](https://pkg.go.dev/math/rand#Source), usually wrapped in a [Rand](https://pkg.go.dev/math/rand#Rand). Both types should be used by a single goroutine at a time: **sharing among multiple goroutines requires some kind of synchronization**.

As I understand generating new random number updates PRNG state and therefore cannot be done without synchronisation .

sesposito: I stand corrected.

VBorisof: Omitted naming

sesposito (approved):

PR #2489, opened 2026-04-06
Refactor leaderboard scheduler to prevent race conditions.

sesposito:
The new design ensures a main loop controls all the scheduling logic, resolving some race conditions which could prevent the hooks from firing in some conditions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,11 @@ The format is based on [keep a changelog](http://keepachangelog.com) and this pr
 ### Added
 - Improve logic in Apple JWS receipts validation.
 - Add `console.disable_display_name_search` config option to disable console account search by display name entirely.
+
 ### Fixed
 - Fix an issue where group admin account deletion could leave orphaned group edges.
 - Fix In-App Purchase validation persist parameter handling for Apple JWS receipts.
+- Fix race conditions in leaderboard/tournament scheduler which could prevent hooks from triggering under certain conditions.
 
 ## [3.39.0] - 2026-05-20
 ### Added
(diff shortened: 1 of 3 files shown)


omer-sequel:
@sesposito will this be part of the next release? it fixes two bugs that I've encountered with tournament resets:

- We have two tournaments which share the reset time, scheduled separately. When the reset time is reached, only the one which was scheduled first is reset
- Reset was called before actual time, presumably due to clock jitter? Left me in a stranded state where no resets were called until a new tournament was scheduled. 

Both on 3.38.0, but it seems like there is no change in the scheduling logic between 3.38.0 and 3.39.0

mofirouz:
> @sesposito will this be part of the next release? it fixes two bugs that I've encountered with tournament resets:
> 
> * We have two tournaments which share the reset time, scheduled separately. When the reset time is reached, only the one which was scheduled first is reset
> * Reset was called before actual time, presumably due to clock jitter? Left me in a stranded state where no resets were called until a new tournament was scheduled.
> 
> Both on 3.38.0, but it seems like there is no change in the scheduling logic between 3.38.0 and 3.39.0

@omer-sequel We are QAing this change; As you can tell, this is a large architectural change to how we process the leaderboards; Some of the games we power have thousands of leaderboards so we need to ensure that this single queue system can scale for that. 

Having said that, if you are on Heroic Cloud, I'm happy to package up an early build of this for you to run and test, on top of the latest Nakama (Enterprise) available. Feel free to ping support@ email address and quote this thread and we can prep this for you. Thanks, Mo.

omer-sequel:
@mofirouz thanks for the response, I can imagine it's quite a major change :) We're self-hosting for now, but I'll hit up support if that were to change 🙏

Calm0 (approved):

Calm0:
LGTM

zyro (approved):

PR #2493, opened 2026-04-17
Prevent Lua/JS storage index filter deadlocks

sesposito:
JS/Lua runtimes storage index filter function require a VM for its execution, if an RPC trigger a filter function call it can create a deadlock.
This PR allows the reuse of the currently checked out VM for an incoming RPC request so that no new JS VM is required, preventing the deadlock.

### DIFF
--- a/server/runtime_javascript.go
+++ b/server/runtime_javascript.go
@@ -45,6 +45,8 @@ import (
 
 const JsEntrypointFilename = "index.js"
 
+type ctxJsRuntimeKey struct{}
+
 type RuntimeJS struct {
 	logger       *zap.Logger
 	node         string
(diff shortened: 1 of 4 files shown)


server/runtime_javascript.go
```diff
@@ -206,6 +208,7 @@ func (rp *RuntimeProviderJS) Rpc(ctx context.Context, id string, headers, queryP
 	}
 
 	ctx = NewRuntimeGoContext(ctx, r.node, r.version, r.envMap, RuntimeExecutionModeRPC, headers, queryParams, traceID, expiry, userID, username, vars, sessionID, clientIP, clientPort, lang)
+	ctx = context.WithValue(ctx, ctxJsRuntimeKey{}, r)
```
redbaron, line 211: This is RPC only. What about before/after hooks and other places where runtime functions can be executed (match loop?)

server/runtime_javascript.go
```diff
@@ -2415,7 +2425,6 @@ func (rp *RuntimeProviderJS) StorageIndexFilter(ctx context.Context, indexName s
 	r.SetContext(ctx)
 	retValue, err, _ := r.InvokeFunction(RuntimeExecutionModeStorageIndexFilter, "storageIndexFilter", fn, jsLogger, nil, nil, "", "", "", nil, 0, "", "", "", "", r.vm.ToValue(objectMap))
 	r.SetContext(context.Background())
```
redbaron, line 2427: r is a runtime module instance running an RPC. I think this will make it resume executing original function with a background context

server/runtime_javascript.go
```diff
@@ -2366,10 +2369,19 @@ func (rp *RuntimeProviderJS) SubscriptionNotificationGoogle(ctx context.Context,
...
+		r, err = rp.Get(ctx)
+		if err != nil {
+			return false, err
+		}
+		defer rp.Put(r)
 	}
+
 	jsFn := r.GetCallback(RuntimeExecutionModeStorageIndexFilter, indexName)
 	if jsFn == "" {
 		rp.Put(r)
```
redbaron, line 2387: is it safe if there was a "parent" RPC?

redbaron:
I think having a test showing Nakama with vm pool size of 1 making multiple indexed writes can help here.

PR #2494, opened 2026-04-21
Update core_authenticate.go

CamelleoDev:
Added a check whether an account with a specific email already exists. If it does, return Codes.AlreadyExists to the client.


novabyte:
@CamelleoDev Thanks for the code contribution.

> If there already is an existing email account, fail. Why was this never added until now?

The reason it was never added is because it opens up an attack vector against the game backend. It's called an account enumeration attack and is considered a low-medium severity issue. The knowledge that someone has an account itself is private information.

CamelleoDev:
Alright, I see what you mean, I’ll see what I could change. I did notice there was an email verification status, assuming that could negate it in some sort of way.

PR #2495, opened 2026-04-24
Add runtime functions to write storage objects with retries.

sesposito:


server/core_storage.go
```diff
@@ -682,6 +685,75 @@ func storageWriteObjects(ctx context.Context, logger *zap.Logger, metrics Metric
...
+			if write.Key == "" {
+				return nil, errors.New("write error: expects key to be a non-empty string")
+			}
+			if write.UserID != "" {
+				if _, err := uuid.FromString(write.UserID); err != nil {
+					return nil, errors.New("write error: expects an empty or valid user id")
+				}
+			}
+			if maybeJSON := []byte(write.Value); !json.Valid(maybeJSON) || bytes.TrimSpace(maybeJSON)[0] != byteBracket {
+				return nil, errors.New("write error: value must be a JSON-encoded object")
```
andresfelipemendez, line 711: can json.Valid(maybeJSON) say where is the invalid json?

andresfelipemendez (approved):

server/core_storage.go
sesposito: It only returns a bool

PR #2502, opened 2026-05-13
Fix regression in X-Forwarded-For header handling.

sesposito:
This issue prevented the origin IP to be correctly forwarded to Satori for ip geolocation of identities created via before/after hooks.


redbaron:
Maybe related: https://linear.app/heroiclabs/issue/MC-2027

novabyte (approved):
@sesposito 👍

Issue #2394, opened 2025-09-26
[Feature Request] Add Variants of Storage API that accept passed in db.Tx


Zhennan-Xu:
When writing complex custom rpcs, it is often desired to have control over the db transaction lifecycle. The storage APIs however start their own transaction; to avoid isolation issues, it would be ideal if the storage APIs could have additional variants that accept a pre-existing transaction.

relevant post:
https://forum.heroiclabs.com/t/feature-request-add-variants-of-storage-api-that-accept-passed-in-db-tx/426

murraystewart96:
Hopefully this helps - https://github.com/heroiclabs/nakama/pull/2511

PR #2511, opened 2026-06-05
2394 Adding storage tx lifecyle management to runtime API

murraystewart96:
https://github.com/heroiclabs/nakama/issues/2394

Implement storage transaction lifecycle across Go, Lua, and JavaScript runtimes

When writing complex RPCs it's often necessary to group multiple storage operations into a single atomic transaction. Previously the storage APIs each managed their own transaction internally, making this impossible.

This adds txBegin/txCommit/txRollback functions to all three runtimes (Go, Lua, JS) and updates storageRead, storageWrite, and storageDelete to accept an optional transaction handle.

All existing call sites are unchanged in behaviour — they pass nil and continue using the internal auto-transaction path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project are documented below.
 The format is based on [keep a changelog](http://keepachangelog.com) and this project uses [semantic versioning](http://semver.org).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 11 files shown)


murraystewart96:
depends on - https://github.com/heroiclabs/nakama-common/pull/226

PR #2513, opened 2026-06-09
Fix/quizverse live banner public event window

intelli-verse-x-health-tech:


mofirouz:
We are permanently blocking Intelliverse and all their employees from opening PRs on Nakama.

PR #2517, opened 2026-06-19
Annotations for non-hook initializer module methods

yinnie:
Before we add the methods to the Function Reference page, we need to add annotations. 
This is part 1. for the non-hooks.  For review before getting the rest. 

https://linear.app/heroiclabs/issue/DX-1327/add-nakama-initializer-module-api-to-function-ref-page


server/runtime_go.go
```diff
@@ -90,21 +94,38 @@ func (ri *RuntimeGoInitializer) GetConfig() (runtime.Config, error) {
...
+// @group events
+// @summary Register a function that is triggered when client sessions end.
+// @param fn(type=function) The function to execute when a client session ends.
+// @return error(error) An optional error value if an error occurred.
 func (ri *RuntimeGoInitializer) RegisterEventSessionEnd(fn func(ctx context.Context, logger runtime.Logger, evt *api.Event)) error {
 	ri.sessionEndFunctions = append(ri.sessionEndFunctions, fn)
 	return nil
 }
 
+// @group registration
```
sesposito: I think rpc is a better group descriptor here.
```suggestion
// @group rpc
```

server/runtime_go.go
```diff
@@ -2734,6 +2763,10 @@ func (ri *RuntimeGoInitializer) RegisterMatchmakerOverride(fn func(ctx context.C
 	return nil
 }
 
+// @group matchmaker
+// @summary Register a custom matchmaker processor function. Receives matchmaker entries and returns groups of entries that should be matched together.
```
sesposito:
```suggestion
// @summary Register a custom matchmaker processor function. Receives entire pool of matchmaker entries and returns groups of entries that should be matched together.
```

server/runtime_go.go
```diff
@@ -2814,10 +2875,25 @@ func (ri *RuntimeGoInitializer) RegisterSubscriptionNotificationGoogle(fn func(c
...
+// @param sortableFields(type=[]string) The fields that can be used for sorting query results.
+// @param maxEntries(type=int) The maximum number of entries to keep in the index.
+// @param indexOnly(type=bool) If true, objects are indexed but not persisted to storage.
+// @return error(error) An optional error value if an error occurred.
 func (ri *RuntimeGoInitializer) RegisterStorageIndex(name, collection, key string, fields []string, sortableFields []string, maxEntries int, indexOnly bool) error {
 	return ri.storageIndex.CreateIndex(context.Background(), name, collection, key, fields, sortableFields, maxEntries, indexOnly)
 }
 
+// @group storage
+// @summary Register a filtering function for a storage index. The filter determines whether a storage object should be included in the index.
```
sesposito:
```suggestion
// @summary Register a filtering function for a storage index. The filter determines whether a storage object should be included or skipped/removed from the index (if present).
```

sesposito (approved):
A few minor suggestions, but otherwise nice work in adding these missing annotations @yinnie 👍

PR #2519, opened 2026-06-23
Fix Apple JWS IAP persist flag

sesposito:


server/core_purchase.go
```diff
@@ -85,30 +85,45 @@ func ValidatePurchasesApple(ctx context.Context, logger *zap.Logger, db *sql.DB,
...
-		Environment:      env,
-		SeenBefore:       dbPurchase.seenBefore,
-	}
+
+	validatedPurchase.UserId = suid
+	validatedPurchase.SeenBefore = dbPurchase.seenBefore
+	validatedPurchase.CreateTime = timestamppb.New(dbPurchase.createTime)
+	validatedPurchase.UpdateTime = timestamppb.New(dbPurchase.updateTime)
+	validatedPurchase.RefundTime = timestamppb.New(dbPurchase.refundTime)
+	validatedPurchase.ProviderResponse = ""
```
Calm0: Do we want this to be empty?

sesposito: This field makes sense for non-JWS receipts so the caller can inspect any further details from the provider response if he needs to, but since JWS validation happens without external requests, returning the receipt back to the caller didn't seem particularly useful, that's why it's being set to empty. Purchases/Subscription listing operations would still return the attached receipt.

Calm0: Ah yes makes sense, maybe a small comment and setting it in the non persist path as well to show that it's intentional?

Calm0 (approved):

PR #2540, opened 2026-08-13
Fix tournament haystack previous cursor

yaosongding:
## Summary

Fixes a copy-paste bug in `TournamentRecordsHaystack` where the **previous cursor** was assigned the value of the **next cursor**:

```go
// before
tournamentRecordList := &api.TournamentRecordList{Records: results.Records, NextCursor: results.NextCursor, PrevCursor: results.NextCursor, RankCount: results.RankCount}
// after
tournamentRecordList := &api.TournamentRecordList{Records: results.Records, NextCursor: results.NextCursor, PrevCursor: results.PrevCursor, RankCount: results.RankCount}
```

## Impact

Clients paginating backward through `ListTournamentRecordsAroundOwner` receive a next-page cursor where a previous-page cursor belongs. Following the returned `prev_cursor` lands on the **next** page (duplicated forward navigation), and when the leaderboard end leaves no next cursor, no previous cursor is returned at all — backward pagination is broken in both cases.

## Evidence

- Introduced in `5666361a` ("Add leaderboard record list around owner cursors", #877, 2022) and present since. The sibling path in the **same file** — `TournamentRecordsList` (`core_tournament.go`) — already maps `PrevCursor: records.PrevCursor` correctly, as does `LeaderboardRecordsHaystack` (`core_leaderboard.go`). Only the haystack wrapper dropped it.
- The source value `results.PrevCursor` has been correct and populated the entire time — only the tournament wrapper dropped it.
- Verified with git history: the line was introduced once and never touched since.

## Test

Adds `TestApiTournamentHaystack` (`server/api_tournament_test.go`), an API-level test that:

1. Creates a non-authoritative tournament with `enable_ranks`
2. Writes 5 records via `WriteTournamentRecord` (scores 10–50, so the owner at score 30 sits at rank 3)
3. Queries `ListTournamentRecordsAroundOwner` from the middle with a small limit (3)
4. Asserts the middle page contains exactly the records around the owner — scores 40, 30, 20 with ranks 2, 3, 4 — and that `PrevCursor != NextCursor`
5. Asserts following `prev_cursor` returns the single record ranked **above** the middle page (score 50, rank 1), and following `next_cursor` returns the single record ranked **below** (score 10, rank 5)
6. Asserts the three pages together cover all five records exactly once

Verified: test FAILS on the buggy code (the cursor-inequality assertion and the previous-page content assertion both fail — the buggy `prev_cursor` returns the score-10 record instead of the score-50 record), PASSES with the fix.

## Verification

- `go build ./server/` — passes
- `golangci-lint run --new-from-rev=master ./server/` — 0 issues
- New test passes with the fix, fails without it (red-green verified)

### DIFF
--- a/server/api_tournament_test.go
+++ b/server/api_tournament_test.go
@@ -1,0 +1,170 @@
+// Copyright 2026 The Nakama Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/gofrs/uuid/v5"
+	"github.com/heroiclabs/nakama-common/api"
+	"github.com/heroiclabs/nakama/v3/apigrpc"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


sesposito (approved):
Hi @yaosongding, thank you for your contribution; LGTM we'll merge it 👍

PR #2542, opened 2026-08-20
Add named multi FleetManager support

andresfelipemendez:
Add `RegisterFleetManagerByName` and `GetFleetManagerByName` to allow for registering multiple fleet managers and get them by key

Existing `RegisterFleetManager` and `GetFleetManager` register it using an empty string

### DIFF
--- a/server/runtime_go.go
+++ b/server/runtime_go.go
@@ -3625,9 +3625,6 @@ func (ri *RuntimeGoInitializer) RegisterFleetManagerByName(name string, fleetMan
 	if fleetManager == nil {
 		return errors.New("fleet manager cannot be nil")
 	}
-	if _, found := ri.fleetManagers[name]; found {
-		return fmt.Errorf("fleet manager %q is already registered", name)
-	}
 	if err := fleetManager.Init(ri.nk, ri.fmCallbackHandler); err != nil {
 		return fmt.Errorf("failed to run fleet manager Init function: %s", err.Error())
 	}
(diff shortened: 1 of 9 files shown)


vendor/github.com/blugelabs/bluge/search/search.go
sesposito, line 1: We'll need to revert this change

server/runtime_go.go
```diff
@@ -3755,6 +3764,8 @@ func NewRuntimeProviderGo(ctx context.Context, logger, startupLogger *zap.Logger
 		matchLock: matchLock,
 
 		fmCallbackHandler: fmCallbackHandler,
+
+		fleetManagers: nk.fleetManagers,
```
sesposito: I think we're setting fleetManagers from `nk.fleetManagers`, but the initializer registers them in `ri.fleetManagers`

server/runtime_go.go
```diff
@@ -3613,16 +3613,25 @@ func (ri *RuntimeGoInitializer) RegisterStorageIndexFilter(indexName string, fn
...
+// @param name(type=string) The name to register the fleet manager under. The empty name is the default.
+// @param fleetManager(type=runtime.FleetManagerInitializer) The fleet manager implementation to register.
+// @return error(error) An optional error value if an error occurred.
+func (ri *RuntimeGoInitializer) RegisterFleetManagerByName(name string, fleetManager runtime.FleetManagerInitializer) error {
 	if fleetManager == nil {
 		return errors.New("fleet manager cannot be nil")
 	}
+	if _, found := ri.fleetManagers[name]; found {
+		return fmt.Errorf("fleet manager %q is already registered", name)
+	}
```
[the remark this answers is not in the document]
sesposito: Just to make it consistent with our other Register* APIs, I think we want to just overwrite if called twice for the same namespace.

sesposito (changes requested):
@andresfelipemendez left a few comments. Can you please also double check if we emit a log when a FM is registered? If not, we should 🙏

server/runtime_go.go
andresfelipemendez: changes to initialize empty map

sesposito (approved):

PR #2551, opened 2026-09-04
fix: use valid WWW-Authenticate header for REST 401 responses

### LINKED ISSUE #1264: Unauthenticated REST response contains invalid `WWW-Authenticate` header
## Description

Unauthenticated REST responses contain an invalid `WWW-Authenticate` header. This is a problem for HTTP clients that implement the [general HTTP authentication framework](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#the_general_http_authentication_framework). In my case, [Dart's `HttpClient`](https://api.dart.dev/stable/3.5.0/dart-io/HttpClient-class.html) chokes on the invalid header value and throws an exception instead of continuing to process the response.

## Steps to Reproduce

1. Start a nakama instance
2. Make an unauthenticated request against the REST API: `curl -v -X POST localhost:7350/v2/session/logout`

## Expected Result

The response should either contain no `WWW-Authenticate` header, or one that adheres to the [specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#syntax), e.g. `Basic realm=<realm>`.

## Actual Result

```bash
curl -v -X POST localhost:7350/v2/session/logout

* Host localhost:7350 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:7350...
* Connected to localhost (::1) port 7350
> POST /v2/session/logout HTTP/1.1
> Host: localhost:7350
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 401 Unauthorized
< Cache-Control: no-store, no-cache, must-revalidate
< Content-Type: application/json
< Vary: Accept-Encoding
< Www-Authenticate: Auth token required
< Date: Wed, 14 Aug 2024 16:26:10 GMT
< Content-Length: 44
< 
* Connection #0 to host localhost left intact
{"code":16, "message":"Auth token required"}%  
```

## Your Environment
- Nakama: 3.23.0

wooyang2018:
## Summary

Fixes #1264

When the REST API returns 401 Unauthenticated, grpc-gateway's default error handler sets `WWW-Authenticate` to the raw gRPC error message (e.g. `Auth token required`), which is not a valid challenge per RFC 7235 / RFC 6750. Some HTTP clients (e.g. Dart HttpClient) fail to parse the response.

This PR registers a custom grpc-gateway error handler that replaces the invalid header with `Bearer realm="nakama"` on 401 responses.

## Changes

- Add `handleHTTPError` and `wwwAuthenticateFixWriter` in `server/api.go`
- Wire `grpcgw.WithErrorHandler(handleHTTPError)` on the API gateway
- Route `handleRoutingError` through the same handler for consistency

## Test plan

- [x] `go test ./server/ -run TestWWWAuthenticateHeaderOnUnauthenticated -v` passes locally
- [x] `go build -trimpath -mod=vendor` passes locally
- [ ] CI integration tests (docker-compose-tests)

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -172,6 +172,7 @@ func StartApiServer(logger *zap.Logger, startupLogger *zap.Logger, db *sql.DB, p
 	// Should start after GRPC server itself because RegisterNakamaHandlerFromEndpoint below tries to dial GRPC.
 	ctx := context.Background()
 	grpcGateway := grpcgw.NewServeMux(
+		grpcgw.WithErrorHandler(handleHTTPError),
 		grpcgw.WithRoutingErrorHandler(handleRoutingError),
 		grpcgw.WithMetadata(func(ctx context.Context, r *http.Request) metadata.MD {
 			// For RPC GET operations pass through any custom query parameters.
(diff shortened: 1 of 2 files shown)


server/api.go
```diff
@@ -825,5 +826,21 @@ func handleRoutingError(ctx context.Context, mux *grpcgw.ServeMux, marshaler grp
...
+}
+
+type wwwAuthenticateFixWriter struct {
+	http.ResponseWriter
+}
+
+func (w *wwwAuthenticateFixWriter) WriteHeader(statusCode int) {
+	if statusCode == http.StatusUnauthorized {
+		w.ResponseWriter.Header().Del("WWW-Authenticate")
+		w.ResponseWriter.Header().Set("WWW-Authenticate", `Bearer realm="nakama"`)
```
sesposito, line 843: I think that if we use Set() here we shouldn't need to call `.Del()` above.

sesposito (changes requested):
Hi @wooyang2018, thank you for your contribution. I've left a minor note, after addressed we'll gladly approve and merge.

sesposito (approved):

Issue #2449, opened 2026-01-12
Issue: Incorrect gRPC status code for storage version check error


epishev-m:
# Issue: Incorrect gRPC status code for storage version check error

## Problem Description

When executing the `WriteStorageObjectsAsync` operation and encountering the error "Storage write rejected - version check failed", the Nakama server returns an incorrect gRPC status code.

## Current Behavior

When the error "Storage write rejected - version check failed" occurs, the following is returned:
- **HTTP status code**: 400
- **gRPC status code**: `3` (INVALID_ARGUMENT)

## Expected Behavior

When the error "Storage write rejected - version check failed" occurs, the following should be returned:
- **HTTP status code**: 400 (or another appropriate code)
- **gRPC status code**: `9` (FAILED_PRECONDITION)

## Rationale

A version check failure is a precondition error, not an invalid argument error. Semantically, this means that the operation cannot be performed due to a state mismatch (object version does not match), which corresponds to the `FAILED_PRECONDITION` status according to the gRPC specification.

## Context

Client code contains a temporary workaround to handle this situation. Due to the incorrect gRPC status code, we are forced to check the error message text against a constant string, as there is no other reliable way to identify this specific error case:

```csharp
// Remove this once the server is updated to return the correct error code
if (statusCode == 400 && grpcStatusCode == 3 &&
    exception.Message == "Storage write rejected - version check failed.")
{
    // Handle version check error
    // ...
}
```

This approach is fragile, as it depends on exact message text matching, which may change. This workaround should be removed after the server is fixed.

## Steps to Reproduce

1. Execute a storage write operation (`WriteStorageObjectsAsync`)
2. Attempt to write an object with an incorrect version (version mismatch)
3. Check the returned gRPC status code in the `ApiResponseException`

## Expected Result After Fix

After the server is fixed:
- gRPC status code should be `9` (FAILED_PRECONDITION)
- The workaround in client code can be removed
- Error handling will occur automatically based on the correct status code

PR #2556, opened 2026-09-13
fix: return RESOURCE_EXHAUSTED for oversized RPC HTTP bodies

### LINKED ISSUE #2450: Issue: Incorrect gRPC status code for "request body too large" error
# Issue: Incorrect gRPC status code for "request body too large" error

## Problem Description

When executing storage write operations in Nakama and encountering the error "http: request body too large", the server returns an incorrect gRPC status code.

## Current Behavior

When the error "http: request body too large" occurs, the following is returned:
- **HTTP status code**: 400 (or another appropriate code)
- **gRPC status code**: `3` (INVALID_ARGUMENT)

## Expected Behavior

When the error "http: request body too large" occurs, the following should be returned:
- **HTTP status code**: 400 (or another appropriate code)
- **gRPC status code**: `8` (RESOURCE_EXHAUSTED)

## Rationale

The "request body too large" error means that the request body size exceeds the allowed limit. This is a resource exhaustion error, not an invalid argument error. Semantically, this means that the operation cannot be performed due to resource constraints (request size exceeds the limit), which corresponds to the `RESOURCE_EXHAUSTED` status according to the gRPC specification.

## Context

Client code may require special handling for this situation. Due to the incorrect gRPC status code, we are forced to check the error message text against a constant string, as there is no other reliable way to identify this specific error case. This approach is fragile, as it depends on exact message text matching, which may change.

## Steps to Reproduce

1. Execute a storage write operation (`WriteStorageObjectsAsync`)
2. Attempt to write an object whose size exceeds the server's allowed limit
3. Check the returned gRPC status code in the `ApiResponseException`

## Expected Result After Fix

After the server is fixed:
- gRPC status code should be `8` (RESOURCE_EXHAUSTED)
- Error handling will occur automatically based on the correct status code
- Client code will be able to reliably identify this error without checking the message text

wooyang2018:
## Summary
- HTTP RPC (`/v2/rpc/...` POST): oversized request bodies now return gRPC status **8 (`RESOURCE_EXHAUSTED`)** in the JSON error payload instead of **3 (`INVALID_ARGUMENT`)** ([#2450](https://github.com/heroiclabs/nakama/issues/2450)).
- Adds a unit test asserting the `requestBodyTooLargeBytes` payload uses code 8.

## Scope note
This PR only covers the **RPC HTTP** path in `api_rpc.go`. Storage write version failures ([#2449](https://github.com/heroiclabs/nakama/issues/2449)) and gRPC-Gateway `/v2/storage` body limits are **not** included here (separate follow-up).

## Test plan
- [x] `go test ./server/ -run TestRequestBodyTooLargeResponseUsesResourceExhaustedCode`

Fixes #2450

### DIFF
--- a/server/api_rpc.go
+++ b/server/api_rpc.go
@@ -41,7 +41,7 @@ var (
 	rpcFunctionNotFoundBytes = []byte(`{"error":"RPC function not found","message":"RPC function not found","code":5}`)
 	internalServerErrorBytes = []byte(`{"error":"Internal Server Error","message":"Internal Server Error","code":13}`)
 	badJSONBytes             = []byte(`{"error":"json: cannot unmarshal object into Go value of type string","message":"json: cannot unmarshal object into Go value of type string","code":3}`)
-	requestBodyTooLargeBytes = []byte(`{"code":3, "message":"http: request body too large"}`)
+	requestBodyTooLargeBytes = []byte(`{"code":8, "message":"http: request body too large"}`)
 )
 
 func (s *ApiServer) RpcFuncHttp(w http.ResponseWriter, r *http.Request) {
--- a/server/api_test.go
+++ b/server/api_test.go
@@ -296,6 +296,17 @@ func UserIDFromSession(session *api.Session) (uuid.UUID, error) {
 	return uuid.FromString(data["uid"].(string))
 }
 
+func TestRequestBodyTooLargeResponseUsesResourceExhaustedCode(t *testing.T) {
+	t.Parallel()
+
+	if !strings.Contains(string(requestBodyTooLargeBytes), `"code":8`) {
+		t.Fatalf("expected ResourceExhausted (8) in RPC body-too-large payload, got %q", string(requestBodyTooLargeBytes))
+	}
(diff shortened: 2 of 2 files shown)


zyro (changes requested):
GRPC Gateway would translate GRPC error code 8 `RESOURCE_EXHAUSTED` into a HTTP `429 Too Many Requests` response code, which is not strictly speaking appropriate here.

If we're making a change for correctness, the most accurate response would be HTTP `413 Content Too Large`.

See mappings here: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/runtime/errors.go#L56

wooyang2018:
Thanks for the review. Updated the RPC HTTP path to return HTTP 413 (StatusRequestEntityTooLarge) for oversized POST bodies. The JSON payload keeps gRPC code 3 (INVALID_ARGUMENT) rather than 8 (RESOURCE_EXHAUSTED), since grpc-gateway maps 8 to HTTP 429, which would be misleading here. Added a unit test asserting 413 and that we do not emit code 8.

PR #2557, opened 2026-09-13
fix: return FAILED_PRECONDITION for storage version conflicts

### LINKED ISSUE #2449: Issue: Incorrect gRPC status code for storage version check error

wooyang2018:
## Summary
- Storage writes that fail optimistic concurrency (`ErrStorageRejectedVersion`) now return gRPC **`FAILED_PRECONDITION` (9)** instead of **`INVALID_ARGUMENT` (3)** ([#2449](https://github.com/heroiclabs/nakama/issues/2449)).
- Permission denials on write remain **`INVALID_ARGUMENT`**.
- Updates `core_storage_test.go` expectations accordingly.

## Test plan
- [x] `go test ./server/ -run 'TestStorageWriteRuntimeGlobalIfMatchNotExists|TestOCCNotExistsAuthoritative|TestOCCWriteSameValueWithOutdatedVersionFail'` (with `TEST_DB_URL` / Postgres)

Fixes #2449

### DIFF
--- a/server/core_storage.go
+++ b/server/core_storage.go
@@ -589,7 +589,10 @@ func StorageWriteObjects(ctx context.Context, logger *zap.Logger, db *sql.DB, me
 		var writeErr error
 		sortedWrites, acks, writeErr = storageWriteObjects(ctx, logger, metrics, tx, authoritativeWrite, ops)
 		if writeErr != nil {
-			if errors.Is(writeErr, runtime.ErrStorageRejectedVersion) || errors.Is(writeErr, runtime.ErrStorageRejectedPermission) {
+			if errors.Is(writeErr, runtime.ErrStorageRejectedVersion) {
+				return StatusError(codes.FailedPrecondition, "Storage write rejected.", writeErr)
+			}
+			if errors.Is(writeErr, runtime.ErrStorageRejectedPermission) {
 				return StatusError(codes.InvalidArgument, "Storage write rejected.", writeErr)
 			}
 			return writeErr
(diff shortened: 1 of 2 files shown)


zyro (approved):
This gives a more meaningful GRPC error code but maps to the same HTTP `400 Bad Request` response code.

